PR #17, opened 2016-11-16
Pkg3: correcting invalid version compatibility claims

simonbyrne:
As pointed out in https://github.com/JuliaLang/Juleps/issues/3#issuecomment-256164888, it is not always possible to fix broken compatibility claims by tagging new versions.

Thoughts?


StefanKarpinski:
This is fine as far as it goes, but it doesn't really spell things out in great detail.

PR #21, opened 2016-12-10
[WIP] Nullable Redesign

johnmyleswhite:
This draft describes the core ideas behind a redesign of nullables. It will require substantial elaboration before we have a final design, but my goal is to get this out in front of people while we are still debating the remaining points of uncertainty in the design. The primary source of remaining uncertainty are the arguments in favor of `Union{T, Void}` vs `Union{Some{T}, Void}`.

### DIFF
--- a/Nullable.md
+++ b/Nullable.md
@@ -1,0 +1,442 @@
+# JULEP Nullable
+
+- **Title:** Nullable Redesign
+- **Authors:** John Myles White <<jmw@johnmyleswhite.com>>
+- **Created:** December 9, 2016
+- **Status:** work in progress
+
+## Abstract
+
+The representation of missing values in Julia is too complicated. We currently
+have at least three kinds of null-like constructs:
+
+* A `Void` type with a singleton value `nothing`, which is the return value of
+    functions that don't return another kind of value;
+* A `Nullable{T}` parametric type with many values, which allows programmers
+    to represent the possibility of missing/invalid values;
+* An `#undef` quasi-value, which reflects the use of null pointers in Julia's
+    implementation and is exposed after allocating memory without
+    initializing that memory.
+
+This Julep describes a proposal to merge these three constructs into one
+construct.
+
+In particular, the proposal is to replace all existing null-like constructs
+with a redesigned version of the current `Nullable` type. This new version of
+`Nullable` will involve both (a) changing the internal implementation of the
+current `Nullable` type and (b) clarifying the semantics of the `Nullable`
+type. Because the name `Nullable{T}` is a little long, we will use the
+notation, `?T`, as a short-hand for `Nullable{T}` throughout this document.
+Making this short-hand a genuine part of Julia's syntax is one part of this
+Julep's proposed design.
+
+## Overview of Goals
+
+The core goals for this Julep are to:
+
+* Simplify the language by replacing many similar null-like constructs with a
+    single null-like construct.
+* Ensure that the resulting representation of `null` is a first-class citizen
+    of the language. In particular, we will remove the second-class citizen
+    status of `#undef`, which is a quasi-value that can be present in
+    uninitialized data structures, but cannot be inserted into those data
+    structures.
+* Clarify how functions can be applied to nullable types by formalizing an
+    extensible protocol for applying methods defined over `T` to values of
+    type `?T`.
+* Make it possible to write optimized code for nullable types without having
+    to call `Core.Inference.return_type` or `promote_type`.
+* Introduce syntactic sugar for nullable types and also introduce special
+    binary operators for working with nullable-typed objects.
+
+The optional goals for this Julep are to:
+
+* Replace the `NullableArray{T}` type with `Array{?T}`, while maintaining
+    the ability to pass the raw masked values as a single block of memory
+    through Julia's FFI in the way that `NullableArray{T}` currently allows
+    passing the `values` field without passing the `isnull` mask.
+
+The prerequisites for this Julep are to:
+
+* Complete optimizations to the Julia compiler to handle `Union{T, Void}` as
+    efficiently as our current `Nullable` type handles the union of the set of
+    non-null values `Nullable{T}(val)` and the singleton set containing the
+    null value `Nullable{T}()`. These compiler changes may include memory
+    layout changes as well as optimization of run-time dispatch among
+    pre-specified types via low-cost branches.
+
+## Step 1: Redesigning the Nullable Type's Internal Implementation
+
+Assuming that compiler optimizations are put in place for union types, the
+first step for this Julep is to replace the concrete type `Nullable{T}` with an
+optimized non-concrete union type. The options for replacing `Nullable{T}`
+that have been proposed are the following union types:
+
+* `Union{T, Void}`
+* `Union{Some{T}, Void}`
+* `Union{T, Null{T}}`
+
+Any newly optimized union type would have the same semantics as union types
+currently have: in particular, a union type is not a concrete type and, as
+a consequence, there are no values of any non-trivial union type. The use of
+a non-concrete type is important both for optimization purposes and for
+clarifying the possible semantics we can define for the new nullable types,
+as we would be replacing a concrete type with a non-concrete type.
+
+Deciding between these three options will require balancing a variety of
+concerns. The arguments for each option are summarized below. We include all
+of them for clarity, although we believe that the arguments on behalf of
+`Union{Some{T}, Void}` are the strongest.
+
+### `Union{T, Void}` Pros & Cons
+
+When a function returns `Union{T, Void}`, the user can immediately apply all
+methods defined on `T` when the value happens to be of type `T`. Allowing users
+to call existing methods without having to unwrap non-null values can be
+convenient.
+
+In this regard, the use of a `Union{T, NAtype}` pattern in DataArrays was
+perceived by many to offer greater convenience than our current `Nullable{T}`
+type provides, although the use of a union type was also the cause of that
+library's historical poor performance because of the compiler's non-optimized
+handling of union types. This Julep assumes that union types will be optimized
+such that poor performance is no longer a consequence of using a union type as
+the implementation of `?T`.
+
+Instead of concerns about inferior performance, one major argument against
+using `Union{T, Void}` is that a `Union{T, Void}` type can only be used
+reliably in code if the methods being called on instances of the union type
+are defined for both the `T` and `Void` sub-types of that union. Going forward,
+this document will refer to these two cases as the non-null and null cases,
+respectively.
+
+Because the `Union{T, Void}` representation uses a union type, it is possible
+for a method to be defined only over a subset of the values in the union. In
+particular, the possibility of directly calling `f(x::T)` on a value of type
+`Union{T, Void}` means that it is possible for code to handle the non-null case
+without being capable of handling the null case. We will refer to situations in
+which code handles only one of the non-null or null cases without handling the
+other case as failing to exhibit totality, a term used because the code does
+not define an execution path for the totality of all possible inputs.
+
+To ensure totality, users must either (a) use functions for which both
+`f(x::T)` and `f(x::Void)` are defined or (b) use higher-order functions
+that can take in a function with a method of the form `f(x::T)` and employ a
+default semantics for null values to handle the `f(x::Void)` case. In
+particular, we will refer to the pattern of extending an `f(x::T)` method to
+handle `f(x::Union{T, Void)` by assuming that `f(x::Void) = nothing` as
+lifting.
+
+The unique concern with `Union{T, Void}` is `f(x::T)` will often be defined
+because it is exists even in the parts of Julia that deal with only
+non-nullable types. When `T` is `Int`, this path is any function on integers.
+Many such functions exist and it is possible that users will assume that
+`f(x::Void)` will also exist because `f(x::Int)` exists.
+
+In DataArrays, the closely related problem of needing definitions of
+`f(x::NAtype) = NA` for all functions `f` was addressed by manually defining
+`f(x::NAtype) = NA` for all functions `f` that commonly occurred in practice.
+Although this piece-meal approach worked effectively in many cases because the
+majority of real code makes use of only a small number of core functions,
+there was always yet another function that to had be defined manually when a
+user found that `f(x::NAtype)` was not defined for their favorite function `f`.
+In this sense, the exact problems introduced by manually vectorizing scalar
+functions in Base Julia were repeated again when adding methods for `NAtype`:
+instead of describing a universal pattern using a generic abstraction, we
+repeated that pattern manually over and over again, while always discovering
+new cases that had been left out.
+
+In addition to requiring repetitive code to provide uniform semantics for
+`f(x::Void)`, there are concerns that the convenience of not having to unwrap
+values may encourage unsafe practices that will not be caught by shallow testing
+practices. For example, if a user defines `f(x::T)`, but not `f(x::Void)`,
+and they only test their code against values of type `T`, they may release
+code that they believe will work for `Union{T, Void}`, but will actually
+raise a `MethodError` at run-time if `nothing` is ever encountered.
+
+That said, if we adopted `typealias Nullable{T} Union{T, Void}` as our new
+implementation of `Nullable{T}`, such concerns about safety could be addressed
+by static analysis tools that check that null values are always handled
+whenever they are allowed by a method's type signature. We might use something
+like the approach taken in
+[Flow for Javascript](https://flowtype.org/docs/nullable-types.html).
+
+Finally, the use of `Union{T, Void}` introduces issues with method specificity
+and the related issue of replacing run-time dispatch with compile-time
+dispatch. In particular, the core problem is that `T <: Union{T, Void}`, so the
+types are not disjoint, leading to specificity conflicts.
+
+To see how the lack of disjointness can introduce problems, imagine that a user
+defines the following two functions:
+
+```
+f(x::Int) = 1
+f(x::Union{Int, Void}) = 2
+```
+
+In this case, the second method, which provides a definition for values of type
+`Int`, will actually never be called on any values of type `Int` because the
+second method definition is strictly less specific than the first method
+definition for values of type `Int`. This can even lead to universally dead
+paths in code that are not obvious at first glance:
+
+```
+f(x::Int) = 1
+f(x::Union{Int, Void}) = isa(x, Void) ? 2 : 3
+```
+
+Here the second part of the branch will never be taken, but the unreachability
+of this branch can only be determined by non-local analysis as it is the
+existence of the method, `f(x::Int) = 1`, that renders a branch in other code
+unreachable. Such dead paths are expected in large Julia codebases and can be
+addressed by static analysis tools, but it may be problematic that this
+representation of `?T` could increase the frequency with which these dead paths
+are defined by accident.
+
+As such, the `Union{T, Void}` approach suggests that one would only need to
+define methods for `f(x::Int)` and `f(x::Void)` in practice, rather than
+the union type. Because the `f(x::Void)` definitions would be trivial in
+many cases, we would still need to use general approaches for lifting.
+
+### `Union{Some{T}, Void}` Pros & Cons
+
+This approach involves wrapping the non-null side of a nullable in a `Some`
+wrapper that allows one to distinguish `T` from `Some{T}`. This small amount of
+indirection has many implications because it ensures that
+`T <: Union{Some{T}, Void}` is not true, breaking many of the patterns we noted
+with `Union{T, Void}`.
+
+For example, because a wrapper is employed, no functions with methods defined
+on `T` will work on objects of type `Union{Some{T}, Void}` unless those methods
+are defined for `f(x::Some{T})` and `f(x::Void)`. For example, the existence of
+`f(x::Int)` does not imply that `f(x::Some{Int})` or `f(x::Void)` will be
+defined.
+
+This limitation can also be framed as a benefit because it ensures that
+multiple methods can co-exist without the specificity concerns raised for
+`Union{T, Void}`.
+
+For example, it would be possible to write code that hits all branches of the
+following method definitions:
+
+```
+f(x::Int) = 1
+f(x::Union{Some{Int}, Void}) = isa(x, Void) ? 2 : 3
+```
+
+To repeat this point in other terms,
+the use of `Some{T}` instead of `T` provides a way of having the `T` and `?T`
+types co-exist without specificity issues, but it simultaneously ensures
+that no existing methods for `T` automatically apply to `?T`. It provides
+arguably cleaner semantics (by making `T` and `?T` disjoint) at the cost of
+making the non-null case for nullable types less convenient to work with.
+
+As such, this implementation of `?T` would require that we push users towards
+the consistent usage of specialized lifting machinery that knows how to handle
+`f(x::Union{Some{T}, Void})` given only knowledge of how `f(x::T)` works. For
+one approach to this problem, see the later section on lifting functions from
+non-nullable types to nullable types.
+
+Because users will have to work with lifted functions, this approach makes it
+more difficult for users to write code that works for the non-null case, but
+does not work for the null case. But the possibility of defining methods for
+only one of `Some{T}` or `Void` means that a totality checker would still be
+needed because users can always define only one of `f(x::Some{T})` and
+`f(x::Void)`. (We should note that the current implementation of `Nullable{T}`
+also does not ensure totality.)
+
+Another point in favor of `Some{T}` is that it is sometimes useful to
+distinguish between two confusable outcomes in cases where a method returns
+something like `Union{Some{Void}, Void}` -- as might happen in a hash table
+implementation that returns `Some(nothing)` when a key is present and
+associated with `nothing` as a value, but returns `nothing` when a key is not
+present. This is another implication of working with disjoint types for `T` and
+`?T`.
+
+On the other hand, the use of `Some{T}` raises issues about how we expect `T`,
+`Some{T}` and `Void` to interact. For example, consider the following code:
+
+```
+function sum(xs::Vector{Union{Some{Int}, Void}})
+    s = 0
+    for x in xs
+        s += x
+    end
+    return x
+end
+```
+
+What type does this return when `xs = ?Int[]`? What type does it return when
+`xs = ?Int[Some(1)]`? What type does it return when `xs = ?Int[nothing]`?
+
+At a minimum, the use of `Some{T}` would require careful thought about
+type conversion and type promotion rules since we would need to define methods
+for combinations like `+(x::Int, y::Some{T})`. Almost surely, we would need to
+allow promoting `T` to `Some{T}` to make common code patterns work without
+much verbosity.
+
+### `Union{T, Null{T}}` Pros & Cons
+
+This final approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
+although it seems to be done only because it is falls out naturally from an
+implementation that uses distinct sentinel values for each nullable type in
+the language. Otherwise, this approach is similar to the `Union{T, Void}` case
+described above, except for one crucial problem that would seem to suggest this
+approach offers few benefits over our current implementation of `Nullable{T}`.
+
+That problem is essentially the problem we already have with `Nullable{T}`:
+writing functions that return concrete types as output generally requires
+information about the return types of any functions used inside of a function
+because one needs to know the return type associated with the non-null case
+even when writing code to handle the null case. Needing to know about the code
+path not taken will be referred to the "counterfactual return type problem" in
+what follows.
+
+To see how this issue arises in practice, consider the following possible
+implementation of `map(f, x::Nullable{T})` given our current implementation of
+`Nullable{T}`:
+
+```
+function map(f, x::Nullable{T})
+    if !isnull(x)
+        return Nullable(f(get(x)))
+    else
+        return Nullable{Core.Inference.return_type(f, (T, ))}()
+    end
+end
+```
+
+For example, ensuring that `map(sqrt, Nullable{Int})` always returns
+`Nullable{Float64}` requires that the null path in the code be aware of the
+return type of `sqrt(x::Int)`. This information is not generally available in
+Julia.
+
+In static languages with arrow types, the counterfactual return type problem is
+not a problem at all because the return type of a function is guaranteed to be
+well-defined by the semantics of the language and guaranteed to be knowable at
+compile-time, so one can refer to properties of `f` without having to call into
+the implementation details of the language. But Julia does not provide such
+guarantees, so type-inference must be invoked to determine information about
+the return type of the counterfactual branch. An alternative approach is to
+use `promote_type` to explicitly register return types, but this is clearly
+problematic as it essentially duplicates a function's declared return type,
+but does so using a mechanism that can be inaccurate.
+
+Note that this problem does not generally occur for the non-null case because
+we are expected to evaluate `f` in that scenario. It is only a problem when we
+need to know the return type of `f`, but must not determine this type by
+any actual evaluation of `f`. Because this problem does not occur in the
+non-null case, it does arise for either `Union{T, Void}` or
+`Union{Some{T}, Void}`. The problem is unique to `Union{T, Null{T}}` and our
+current `Nullable{T}` implementation as one must determine the type parameter
+`T` even for the null case.
+
+## Step 2:: Define Core Operations on Nullable Types by Lifting
+
+Once an implementation of `Nullable{T}` is agreed upon, the core semantic issue
+for nullable types is maximizing code reuse by making lifting easy to use. In
+other words, we want to make it easy to evaluate `f(x::Nullable{T})` without
+requiring users to define methods for every function to handle the null case,
+for which the semantics are so consistent. We want to avoid requiring that
+users write code for the null case because we know that the overwhelming
+majority of functions will return `null` as their output whenever any of
+their arguments are `null`.
+
+As mentioned earlier, extending the definition of functions from `T` to `?T`
+is called lifting. We propose the following extensible approach to implementing
+lifting that defaults to the semantics we expect will occur most commonly, but
+allows implementing alternative semantics where needed.
+
+First, we define `broadcast(f, x::Nullable)` as follows:
+
+```
+immutable Lifted{T}
+    func::T
+end
+
+lift(f) = Lifted(f)
+
+broadcast(f, x::Nullable) = broadcast(lift(f), x)
+
+broadcast(f′::Lifted, x::Nullable) = hasvalue(x) ? f′.func(value(x)) : null
+```
+
+This allows us to give default semantics to all functions that implement the
+natural approach to lifting, while still making it possible to handle special
+case functions that require distinct semantics.
+
+Some such functions are:
+
+* `isnull(x::?T)`
+* `get(x::?T)`
+* `get(x::?T, default::T)`
+
+Because we would use broadcasting, lifting would happen via dot-syntax:
+for example, `x .+ y` would be lifted by default. In addition, we would special
+several operators like `+`, `-`, `==`, etc. to automatically be equivalent to
+broadcasting on nullables.
+
+## Step 3: Introduce Syntactic Sugar for Nullable Types
+
+Following the lead of
+[Flow for Javascript](https://flowtype.org/docs/nullable-types.html), we
+would introduce `?T` as sugar for `Nullable{T}`. We have also considered C#'s
+`T?`. Which is chosen will likely depend upon how they interact with the
+ternary operator.
+
+We might also introduce a null-coalescing operator, `??`, with
+right-associativity that essentially performs lifting: for example,
+`x ?? y ?? 0` would be equivalent to `hasvalue(x) ? x : (hasvalue(y) ? y : 0)`.
+
+Finally, we might introduce `.?` for lifted field access such that
+`x?.field_name` would be equivalent to `getfield.(x, :field_name)`.
+
+## Step 4: All Special Operators
+
+There are several special operators that need to be defined. In particular,
+we want to distinguish between non-short-circuiting Booleans and
+short-circuiting Booleans. Following C#'s lead, we will not allow
+short-circuiting Boolean operators to be applied to nullable types.
+
+## Step 5: Memory Initialization and Nullables
+
+At present, `#undef` occurs whenever memory is allocated for mutable types
+without initializing that memory. Thus in the current version of Julia, the
+following occurs:
+
+```
+julia> Array(String, 1)
+1-element Array{String,1}:
+ #undef
+```
+
+Our proposal would remove this behavior entirely.
+
+Instead, there would be two possibilities:
+
+* `Array{String}` can never contain `#undef` values and therefore must be
+    initialized whenever it is allocated.
+* `Array{?String}` is always initialized by the mere act of allocation, as
+    the implementation will default to setting the value of `null` in all
+    positions in the array.
+
+This would likely occur concurrently with an attempt to use `calloc`
+everywhere instead of `malloc`.
+
+## Step 6: Optimizations
+
+When working with nullables, we generally need to check a mask before we
+can decide the value of an operation on nullable typed objects. This check
+can sometimes induce a branch that is costly relative to the cost of the core
+operation being performed (e.g. addition) and the existence of the branch may
+also impede compiler optimizations such as automatic SIMD vectorization.
+
+As such, we will want to ensure that `lift(f, args)` is customized to use
+`ifelse` vs `if` for some functions like addition.
+
+## Extended Usage Examples
+
+* `mean(v::Array{?Int})`
+* `median(v::Array{?Int})`
+* `pdf(Binomial(n, 0.5), x)` where `n::?Int` and `x::?Int`
--- a/Nullable.md
+++ b/Nullable.md
@@ -1,0 +1,191 @@
+# JULEP Nullable
+
+- **Title:** Nullable Redesign
+- **Authors:** John Myles White <<jmw@johnmyleswhite.com>>
+- **Created:** December 9, 2016
+- **Status:** work in progress
+
+## Abstract
+
+The representation of null-like constructs in Julia is too complicated. We
+currently have all of the following kinds of null-like constructs:
+
+* A ` Void` type with a singleton value `nothing`,
+* A `Nullable` type,
+* An `#undef` quasi-value that reflects the use of null pointers in Julia's
+    implementation.
+
+This Julep describes a proposal to coalesce all of these concepts into one.
+
+The proposal is to replace everything with a new version of the `Nullable`
+type. This new version of `Nullable` will involve both changing the
+implementation of our `Nullable` type and clarifying the semantics of this
+type by laying out a thorough description of how the type would be used in
+practice.
+
+The core goals for this Julep are to:
+
+* Simplify the language by replacing many similar constructs with a single
+    construct.
+* Ensure that all representations of `null` are first-class citizens of the
+    language. In particular, we will resolve the problems with `#undef`, which
+    is a quasi-value that can be present in types, but cannot be inserted into
+    them.
+* Resolve the lifting problem for `Nullable` types by formalizing
+    general-purpose, extensible machinery for lifting.
+* Make it possible to write optimized code for `Nullable` types that does not
+    call into type inference via `return_types` or require the use of
+    `promote_type`.
+
+The optional goals for this Julep are to:
+
+* Replace the `NullableArray` type with `Array{Nullable}`, while maintaining
+    the ability to pass the raw values storage unit via FFI in the way that
+    `NullableArray` currently allows.
+
+The prerequisites for this Julep are the following:
+
+* Optimizations to the Julia compiler to handle `Union{S, T}` as efficiently as
+    our current `Nullable` type handles the union of `S` and `Void`. These
+    compiler changes may include both memory layout changes and optimization
+    of run-time dispatch among K pre-specified types into simple branches.
+
+## Step 1: Redefining the Nullable Type
+
+Assuming that compiler optimizations are put in place for union types, the
+first step for this Julep is to replace the concrete type `Nullable` with an
+optimized union type. The options for replacing `Nullable{T}` are:
+
+* `Union{T, Void}`
+* `Union{Some{T}, Void}`
+* `Union{T, Null{T}}`
+
+The `Union` type would have the same semantics as unions currently have. Their
+implementation in the scalar case would likely match that of a possible
+`TaggedUnion` type, which represents a discriminated union, but is a concrete
+type that is not a parent of `T`, `Void`, etc.
+
+Deciding between these options requires balancing a variety of concerns. As
+we currently understand them, the arguments are:
+
+### `Union{T, Void}` Pros & Cons
+
+When a function returns `Union{T, Void}`, the user can immediately apply all
+methods defined on `T` without any unwrapping. This is very convenient and
+does not require the introduction of specialized methods.
+
+The counter-argument is that such code will only work if the methods being
+called are defined for both `T` and `Void`. As such, one must redefine
+`f(x::Void) = nothing` for every new function, even though the semantics are
+identical. The requirement for always defining such repetitive functions
+therefore may be perceived as less safe because code that is tested only
+against values of type `T` may fail in untested cases with a method error if
+applied to a value of type `Void`. This could be resolved by static analysis
+tools that check for totality.
+
+### `Union{Some{T}, Void}` Pros & Cons
+
+This approach is essentially the inverse of the `Union{T, Void}`: here no
+methods will work on objects of this type unless they are defined for both
+`f(x::Some{T})` and `f(x::Void)`. As such, this approach requires consistent
+of specialized lifting machinery via dot-broadcasting syntax. See the later
+section on lifting for details.
+
+Because users will have to work with lifted functions, this approach is
+less convenient, but more general and often safer. It also allows one to
+distinguish between outcomes in cases where a method returns something like
+`Union{Some{Void}, Void}` -- as might happen in a hash table implementation
+that returns `Some` when a key is present and associated with `nothing` as a
+value, but returns `Void` when a key is not present.
+
+### `Union{T, Null{T}}` Pros & Cons
+
+This approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
+although it seems to be done only because it is falls out naturally from an
+implementation that uses sentinel values. Otherwise, it is similar to the
+`Union{T, Void}` case described above, except for one crucial problem.
+
+That problem is essentially the problem we already have with `Nullable{T}`:
+writing "clean" functions requires information about type-inference because of
+a coupling of result types to knowledge about counterfactual outcomes in
+branches.
+
+To see the issue, consider the following possible implementation of
+`map(f, x::Nullable{T})` in our current implementation of `Nullable{T}`:
+
+```
+function map(f, x::Nullable{T})
+    if hasvalue(x)
+        return Nullable(f(value(x)))
+    else
+        return Nullable{return_types(f, T)}()
+    end
+end
+```
+
+The problem is that, when given a null input, the return type can only match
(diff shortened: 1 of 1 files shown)


Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+against values of type `T` may fail in untested cases with a method error if
+applied to a value of type `Void`. This could be resolved by static analysis
+tools that check for totality.
+
+### `Union{Some{T}, Void}` Pros & Cons
+
+This approach is essentially the inverse of the `Union{T, Void}`: here no
+methods will work on objects of this type unless they are defined for both
+`f(x::Some{T})` and `f(x::Void)`. As such, this approach requires consistent
+of specialized lifting machinery via dot-broadcasting syntax. See the later
```
quinnj: "consistent of"?

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+- **Authors:** John Myles White <<jmw@johnmyleswhite.com>>
+- **Created:** December 9, 2016
+- **Status:** work in progress
+
+## Abstract
+
+The representation of null-like constructs in Julia is too complicated. We
+currently have all of the following kinds of null-like constructs:
+
+* A ` Void` type with a singleton value `nothing`,
```
ararslan: Extra space after the backtick

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+branch.
+
+Note that this problem does not occur for the non-null case because we are
+expected to evaluate `f`. It is only a problem when we need to know the return
+type of `f`, but must not determine this type by evaluation of `f`.
+
+## Step 2:: Define Core Operations on Nullable Types by Lifting
+
+The core semantic issue for nullable types is maximizing code reuse: we want
+to evaluate `f(x::?T)`, but not require that every function be defined for
```
ararslan: Should mention what `?T` is intended to mean—it's used here mid-discussion with no prior mention of the syntax

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+This allows us to give default semantics to all functions that implement the
+natural approach to lifting, while still making it possible to handle special
+case functions that require distinct semantics.
+
+Some such functions are:
+
+* `hasvalue`
+* `isnull`
+* `get`
+* `get_or_default`
```
ararslan: I don't think this is a function in Base?

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Step 3: Introduce Syntactic Sugar for Nullable Types
+
+Following the lead of
+[Flow for Javascript](https://flowtype.org/docs/nullable-types.html), we
+would introduce `?T` as sugar for `Nullable{T}`. We have also considered C#'s
+`T?`. Which is chosen will likely depend upon how they interact with the
+ternary operator.
+
+We might also introduce a null-coalescing operator of `??` with
```
ararslan: Sentence flows better without "of" IMO

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+The core semantic issue for nullable types is maximizing code reuse: we want
+to evaluate `f(x::?T)`, but not require that every function be defined for
+both `f(x::T)` and `f(x::?T)` separately as the overwhelming majority of
+functions will return `null` as their output whenever any of their arguments
+are `null`. Implementing this lifting of functions can be solved as follows:
+
+First, we define `broadcast(f, x::Nullable)` as follows:
+
+```
+immutable Lifted{T}
```
ararslan, line 354: Could specify `T<:Function`, since type constructors can't lift (i.e. `Lifted{Int}` whaaat)

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+optimized union type. The options for replacing `Nullable{T}` are:
+
+* `Union{T, Void}`
+* `Union{Some{T}, Void}`
+* `Union{T, Null{T}}`
+
+The `Union` type would have the same semantics as unions currently have. Their
+implementation in the scalar case would likely match that of a possible
+`TaggedUnion` type, which represents a discriminated union, but is a concrete
+type that is not a parent of `T`, `Void`, etc.
```
nalimilan: Maybe explain what would be the advantage of it not being a parent of these types?

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+### `Union{T, Void}` Pros & Cons
+
+When a function returns `Union{T, Void}`, the user can immediately apply all
+methods defined on `T` without any unwrapping. This is very convenient and
+does not require the introduction of specialized methods.
+
+The counter-argument is that such code will only work if the methods being
+called are defined for both `T` and `Void`. As such, one must redefine
+`f(x::Void) = nothing` for every new function, even though the semantics are
```
nalimilan: Could mention the possibility of having a generic fallback definition like this for any function `f` (even if this might be undesirable and Jeff wouldn't support it).

Nullable.md
nalimilan: Probably shouldn't mention dot-broadcasting syntax here: that's one of the possibilities for lifting, but that choice is kind of orthogonal to the issues tackled in this section.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+### `Union{Some{T}, Void}` Pros & Cons
+
+This approach is essentially the inverse of the `Union{T, Void}`: here no
+methods will work on objects of this type unless they are defined for both
+`f(x::Some{T})` and `f(x::Void)`. As such, this approach requires consistent
+of specialized lifting machinery via dot-broadcasting syntax. See the later
+section on lifting for details.
+
+Because users will have to work with lifted functions, this approach is
+less convenient, but more general and often safer. It also allows one to
```
nalimilan: In what sense is it more general?

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+distinguish between outcomes in cases where a method returns something like
+`Union{Some{Void}, Void}` -- as might happen in a hash table implementation
+that returns `Some` when a key is present and associated with `nothing` as a
+value, but returns `Void` when a key is not present.
```
nalimilan: `Some` -> `Some(nothing)`
`Void` -> `nothing`

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+To see the issue, consider the following possible implementation of
+`map(f, x::Nullable{T})` in our current implementation of `Nullable{T}`:
+
+```
+function map(f, x::Nullable{T})
+    if hasvalue(x)
+        return Nullable(f(value(x)))
+    else
+        return Nullable{return_types(f, T)}()
```
nalimilan: `return_types` is actually `Core.Inference.return_type`. The full name isn't required, but the plural could be confusing.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+less convenient, but more general and often safer. It also allows one to
+distinguish between outcomes in cases where a method returns something like
+`Union{Some{Void}, Void}` -- as might happen in a hash table implementation
+that returns `Some` when a key is present and associated with `nothing` as a
+value, but returns `Void` when a key is not present.
+
+### `Union{T, Null{T}}` Pros & Cons
+
+This approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
```
nalimilan, line 281: Can we find cases where this is indeed useful? I guess it could help enforcing type stability in corner cases where a function would get only `nothing` without any type information and would need to return more than `nothing`, e.g. an array which could hold either `nothing` or non-missing values of the (unknown) type `T`. Not sure whether this can happen in practice.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+```
+immutable Lifted{T}
+    func::T
+end
+
+lift(f) = Lifted(f)
+
+broadcast(f, x::Nullable) = broadcast(lift(f), x)
+
+broadcast(f′::Lifted, x::Nullable) = hasvalue(x) ? f′.func(value(x)) : null
```
nalimilan, line 362: `f'` isn't a valid identifier. Also, it would probably work better to overload `Lifted` calls than `broadcast`, to avoid ambiguities.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+This allows us to give default semantics to all functions that implement the
+natural approach to lifting, while still making it possible to handle special
```
nalimilan, line 366: approach -> definition or semantics? The former sounds like a lifting system.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+* `hasvalue`
+* `isnull`
+* `get`
+* `get_or_default`
+
+There are natural extensions of this logic to the n-ary function case.
+
+Because we would use broadcasting, lifting would happen via dot-syntax:
+`x .+ y` would be lifted by default. In addition, we would special several
```
nalimilan: special -> specialize

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+operators like `+`, `-`, `==`, etc. to automatically be equivalent to
+broadcasting on nullables.
```
nalimilan, line 378: Like in C# (for the first two at least).

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Step 3: Introduce Syntactic Sugar for Nullable Types
+
+Following the lead of
+[Flow for Javascript](https://flowtype.org/docs/nullable-types.html), we
+would introduce `?T` as sugar for `Nullable{T}`. We have also considered C#'s
+`T?`. Which is chosen will likely depend upon how they interact with the
```
nalimilan, line 385: Also used by Swift.

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+ternary operator.
+
+We might also introduce a null-coalescing operator of `??` with
+right-associativity that essentially performs lifting: in this case,
+`x ?? y ?? 0` would be equivalent to `hasvalue(x) ? x : (hasvalue(y) ? y : 0)`.
```
nalimilan, line 390: `hasvalue` -> `!isnull`

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+broadcast(f′::Lifted, x::Nullable) = hasvalue(x) ? f′.func(value(x)) : null
+```
+
+This allows us to give default semantics to all functions that implement the
+natural approach to lifting, while still making it possible to handle special
+case functions that require distinct semantics.
+
+Some such functions are:
+
+* `hasvalue`
```
nalimilan: This one doesn't exist in Base either.

Nullable.md
johnmyleswhite: -> "consistent usage of"

Nullable.md
johnmyleswhite: Fixed

Nullable.md
johnmyleswhite: I did this for C# comparisons. This is our two argument `get`. Will indicate that.

Nullable.md
johnmyleswhite: I think we need type constructors to lift if you want things like `pdf(Binomial(n, 0.5), x)` to work correctly when evaluated against a stream of tuples.

Nullable.md
johnmyleswhite: I'm going to remove this section as I think it adds complexity without many gains.

Nullable.md
johnmyleswhite: Agreed

Nullable.md
johnmyleswhite: That's f-prime, which is a valid identifier. What ambiguities do you have in mind?

Nullable.md
ararslan: Oh you're right, I didn't think of that.

Nullable.md
nalimilan: Ah, sorry, it looks the same here. I don't have specific ambiguities in mind, but overloading call would work everywhere and could even be useful on its own, so why not use that?

Nullable.md
ararslan: Or `isnull` and switch the order

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+This approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
+although it seems to be done only because it is falls out naturally from an
+implementation that uses sentinel values. Otherwise, it is similar to the
+`Union{T, Void}` case described above, except for one crucial problem.
+
+That problem is essentially the problem we already have with `Nullable{T}`:
+writing "clean" functions requires information about type-inference because of
+a coupling of result types to knowledge about counterfactual outcomes in
+branches.
```
andyferris: Is it true that this applies? Wouldn't you use just use dispatch, and define e.g. `&(::Bool, Null{Bool})`?

For containers of `Union{T,Null{T}}`, functions like `map` and `broadcast` will have this reliance on inference just as containers of other types always do currently (it won't matter at all that the element type is `Union{T,Null{T}}` as opposed to anything else).

Nullable.md
andyferris: I feel this approach is reminiscent of `NA` from `DataArrays`, but much more precise, in that you can have `true & Null{String}` give an error, say, while `true & Null{Bool}` can return `Null{Bool}`

Nullable.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Step 2:: Define Core Operations on Nullable Types by Lifting
+
+The core semantic issue for nullable types is maximizing code reuse: we want
+to evaluate `f(x::?T)`, but not require that every function be defined for
+both `f(x::T)` and `f(x::?T)` separately as the overwhelming majority of
+functions will return `null` as their output whenever any of their arguments
+are `null`. Implementing this lifting of functions can be solved as follows:
+
+First, we define `broadcast(f, x::Nullable)` as follows:
```
andyferris, line 351: Wait, wait, I'm lost here. If we literally replace `Nullable{T}` with `Union{T, Void}` or `Union{T, Null{T}}` then dispatch will handle function calls. `broadcast`, like *every* other function, would receive a concrete type like `T` or `Void` or `Null{T}`. 

Further, if you mean `typealias Nullable{T} Union{T, Void}`, then we would also have `T <: Nullable` for all possible types `T`.

Like `DataArrays` and `NA`, lifting is not necessary. You **do** need a bunch more method definitions for `Void` / `Null{T}` / `Some{T}`.

andyferris (commented):
Thanks for writing this up, @johnmyleswhite.

I realise this is WIP, but in general I'm left a bit confused whether you want `Nullable` to be a one-element container similar to `RefValue{Union{T,Void}}` or whether you literally mean `typalias Nullable{T} Union{T,Void}` and to use `Void` like `NA` was used in `DataArrays`, except that the compiler will be faster.

Because in the latter case, you don't want to use broadcast and higher-order lifiting, but instead you need method specializations (because methods can only ever receive concrete types).

andyferris:
> my goal is to get this out in front of people while we are still debating the remaining points of uncertainty in the design. The primary source of remaining uncertainty are the arguments in favor of Union{T, Void} vs Union{Some{T}, Void}.

I'm taking this as requests for feedback/opinions (apologies if I misunderstood).

I would advocate strongly for `Union{T, Null{T}}` because I feel it would help *enormously* with writing generic methods. If my method receives a `Null{Vector{Int}}` at least I know what's the element type of the vector that I was supposed to get, and that might help me to arrive at the correct type of my output. Having `Void` here doesn't help me predict what the user wants me to do, and it doesn't help me to write different methods of a function that might have slightly different semantics depending on input types.

I also feel that `Void` and `nothing` already have a meaning, in terms they are the return value of the thing the compiler doesn't have a return value for - like the result of a `for` loop, or an empty function body, or an `if false` statement with no `else` branch. Having to diagnose whether an unfamiliar function returns `nothing` on purpose or just because it just happened to be a "null" value for my chosen input would add unnecessary complication.

I see you also mentioned `#undef`. How would this work? Ideally, I guess we could make the GC aware of how `Union` works. This could be pretty compelling... but we probably need to solve the below:

There is one thing that remains to confuse me. Currently we can have `Nullable{Any}`. However `Union{Any, Void} = Any` (and so-on). I don't think we want to change these properties. So what, for example, would a `Nullable{Any}()` or an `#undef` of `Any` become?

johnmyleswhite:
Thanks for the comments, Andy!

I should point out that the `Union{T, Null{T}}` implementation is not really seriously being considered as far as I understand the growing consensus. Instead, that implementation is being documented to demonstrate why it offers few benefits over our current implementation.

On that note, I think you're misunderstanding the way `Union{T, Void}` would work. That's really helpful for me to see because I'll try harder to ensure that revisions of this Julep communicate this point more clearly.

If we use `Union{T, Void}`, we'll be abandoning the zero-or-one element container way of thinking of `Nullable{T}` and instead we'd be using what is essentially a tagged union that always contains exactly one value: either a value of type `T` or a value of type `Void`. We can think of this as a container to make the use of `broadcast` more reasonable, but we're mostly taking about `broadcast` because it provides minimal syntax for lifting.

Will address your other comments tomorrow morning.

andyferris:
Thanks @johnmyleswhite - that clarifies things immensely. So we would have something which roughly has this structure:
```julia
immutable Nullable{T}
    value::Union{T,Void}
end
size(::Nullable) = ()
getindex(n::Nullable) = n.value
eltype{T}(::Nullable{T}) = Union{T,Void} # ??
isnull(n::Nullable) = isa(n.value, Void)
```
and then use `broadcast`, "lifting", and other machinery to push the nullables through the system?

andyferris:
> I should point out that the Union{T, Null{T}} implementation is not really seriously being considered as far as I understand the growing consensus. Instead, that implementation is being documented to demonstrate why it offers few benefits over our current implementation.

Of course this:
```julia
immutable Nullable{T}
    value::Union{T,Null{T}}
end
```
offers few benefits, and if that is what you meant, then I agree entirely. To be clear, I was suggesting to use "bare" nullables with no wrapper type, to replace `DataArrays.NA` with `Base.Null{T}`, and to remove `Base.Nullable` entirely (except maybe as a typealias for a `Union`).

johnmyleswhite:
Revised the document. Still a bunch more to do (and a few comments left to respond to), but I'm now almost completely convinced we'd want to make `?T` mean `Union{Some{T}, Void}` since the benefits from having `T` and `?T` represent disjoint sets seem to outweigh other considerations.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+To ensure totality, users must either (a) use functions for which both
+`f(x::T)` and `f(x::Void)` are defined or (b) use higher-order functions
+that can take in a function with a method of the form `f(x::T)` and employ a
+default semantics for null values to handle the `f(x::Void)` case. In
+particular, we will refer to the pattern of extending an `f(x::T)` method to
+handle `f(x::Union{T, Void)` by assuming that `f(x::Void) = nothing` as
+lifting.
+
+The unique concern with `Union{T, Void}` is `f(x::T)` will often be defined
+because it is exists even in the parts of Julia that deal with only
```
ararslan, line 131: "it is exists"

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+which code handles only one of the non-null or null cases without handling the
+other case as failing to exhibit totality, a term used because the code does
+not define an execution path for the totality of all possible inputs.
+
+To ensure totality, users must either (a) use functions for which both
+`f(x::T)` and `f(x::Void)` are defined or (b) use higher-order functions
+that can take in a function with a method of the form `f(x::T)` and employ a
+default semantics for null values to handle the `f(x::Void)` case. In
+particular, we will refer to the pattern of extending an `f(x::T)` method to
+handle `f(x::Union{T, Void)` by assuming that `f(x::Void) = nothing` as
```
ararslan, line 127: Missing a `}`

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+On the other hand, the use of `Some{T}` raises issues about how we expect `T`,
+`Some{T}` and `Void` to interact. For example, consider the following code:
+
+```
+function sum(xs::Vector{Union{Some{Int}, Void}})
+    s = 0
+    for x in xs
+        s += x
+    end
+    return x
```
ararslan, line 265: `return s`?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+Note that this problem does not generally occur for the non-null case because
+we are expected to evaluate `f` in that scenario. It is only a problem when we
+need to know the return type of `f`, but must not determine this type by
+any actual evaluation of `f`. Because this problem does not occur in the
+non-null case, it does arise for either `Union{T, Void}` or
+`Union{Some{T}, Void}`. The problem is unique to `Union{T, Null{T}}` and our
+current `Nullable{T}` implementation as one must determine the type parameter
+`T` even for the null case.
+
+## Step 2:: Define Core Operations on Nullable Types by Lifting
```
ararslan, line 335: `2::` -> `2:`?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+would introduce `?T` as sugar for `Nullable{T}`. We have also considered C#'s
+`T?`. Which is chosen will likely depend upon how they interact with the
+ternary operator.
+
+We might also introduce a null-coalescing operator, `??`, with
+right-associativity that essentially performs lifting: for example,
+`x ?? y ?? 0` would be equivalent to `hasvalue(x) ? x : (hasvalue(y) ? y : 0)`.
+
+Finally, we might introduce `.?` for lifted field access such that
+`x?.field_name` would be equivalent to `getfield.(x, :field_name)`.
```
ararslan, line 393: You first mention `.?` but then use `x?.`. Which one is being considered? Or are both?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+short-circuiting Boolean operators to be applied to nullable types.
+
+## Step 5: Memory Initialization and Nullables
+
+At present, `#undef` occurs whenever memory is allocated for mutable types
+without initializing that memory. Thus in the current version of Julia, the
+following occurs:
+
+```
+julia> Array(String, 1)
```
ararslan, line 409: That syntax is "deprecated" per the manual (though no warning or anything is emitted). In the now-preferred syntax, it would be `Array{String,1}(1)`.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+Any newly optimized union type would have the same semantics as union types
+currently have: in particular, a union type is not a concrete type and, as
+a consequence, there are no values of any non-trivial union type. The use of
+a non-concrete type is important both for optimization purposes and for
+clarifying the possible semantics we can define for the new nullable types,
+as we would be replacing a concrete type with a non-concrete type.
+
+Deciding between these three options will require balancing a variety of
+concerns. The arguments for each option are summarized below. We include all
+of them for clarity, although we believe that the arguments on behalf of
```
nalimilan, line 88: "in favor of" rather?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+
+To ensure totality, users must either (a) use functions for which both
+`f(x::T)` and `f(x::Void)` are defined or (b) use higher-order functions
+that can take in a function with a method of the form `f(x::T)` and employ a
+default semantics for null values to handle the `f(x::Void)` case. In
+particular, we will refer to the pattern of extending an `f(x::T)` method to
+handle `f(x::Union{T, Void)` by assuming that `f(x::Void) = nothing` as
+lifting.
+
+The unique concern with `Union{T, Void}` is `f(x::T)` will often be defined
```
nalimilan, line 130: I wouldn't say it's the unique concern. The strongest concern for me is that this nullable type wouldn't allow distinguishing between no value and a stored null value when indexing a dictionary.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+`f(x::NAtype) = NA` for all functions `f` that commonly occurred in practice.
+Although this piece-meal approach worked effectively in many cases because the
+majority of real code makes use of only a small number of core functions,
+there was always yet another function that to had be defined manually when a
+user found that `f(x::NAtype)` was not defined for their favorite function `f`.
+In this sense, the exact problems introduced by manually vectorizing scalar
+functions in Base Julia were repeated again when adding methods for `NAtype`:
+instead of describing a universal pattern using a generic abstraction, we
+repeated that pattern manually over and over again, while always discovering
+new cases that had been left out.
```
nalimilan, line 147: This problem of manual redefinitions is not a criticism of the `Union{T, Void}` approach per se: it would rather justify hard-coding something like (pseudo-code) `(::F){F<:Function}(args...) = nothing` for all functions. (The ~~next~~ third paragraph below gives reasons why this would not be a good idea, but it's a different argument IMHO.)

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+associated with `nothing` as a value, but returns `nothing` when a key is not
+present. This is another implication of working with disjoint types for `T` and
+`?T`.
+
+On the other hand, the use of `Some{T}` raises issues about how we expect `T`,
+`Some{T}` and `Void` to interact. For example, consider the following code:
+
+```
+function sum(xs::Vector{Union{Some{Int}, Void}})
+    s = 0
```
nalimilan, line 261: Should this  be `Some(0)`?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+    for x in xs
+        s += x
+    end
+    return x
+end
+```
+
+What type does this return when `xs = ?Int[]`? What type does it return when
```
nalimilan, line 269: Could you illustrate the alternative solutions? It's not completely clear to me what problem you're describing.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+Julia.
+
+In static languages with arrow types, the counterfactual return type problem is
+not a problem at all because the return type of a function is guaranteed to be
+well-defined by the semantics of the language and guaranteed to be knowable at
+compile-time, so one can refer to properties of `f` without having to call into
+the implementation details of the language. But Julia does not provide such
+guarantees, so type-inference must be invoked to determine information about
+the return type of the counterfactual branch. An alternative approach is to
+use `promote_type` to explicitly register return types, but this is clearly
```
nalimilan, line 322: `promote_op`? Anyway, I don't think it's worth mentioning as it's clearly not a solution.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+
+### `Union{T, Null{T}}` Pros & Cons
+
+This final approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
+although it seems to be done only because it is falls out naturally from an
+implementation that uses distinct sentinel values for each nullable type in
+the language. Otherwise, this approach is similar to the `Union{T, Void}` case
+described above, except for one crucial problem that would seem to suggest this
+approach offers few benefits over our current implementation of `Nullable{T}`.
```
nalimilan, line 286: Should mention possible benefits which justify detailing this approach at all. One benefit I can see is preventing things like `Null{String}() + 1`, which would most likely occur due to a bug somewhere.

What benefits we expect also matters for discussing the problems with this approach. In particular, when inference fails, we could fall back to `Null{Any}` or `Null{Union{}}`. Depending on what that type information is used for, it could be a problem or not.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+case functions that require distinct semantics.
+
+Some such functions are:
+
+* `isnull(x::?T)`
+* `get(x::?T)`
+* `get(x::?T, default::T)`
+
+Because we would use broadcasting, lifting would happen via dot-syntax:
+for example, `x .+ y` would be lifted by default. In addition, we would special
```
nalimilan, line 376: Missing word.

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+the return type of the counterfactual branch. An alternative approach is to
+use `promote_type` to explicitly register return types, but this is clearly
+problematic as it essentially duplicates a function's declared return type,
+but does so using a mechanism that can be inaccurate.
+
+Note that this problem does not generally occur for the non-null case because
+we are expected to evaluate `f` in that scenario. It is only a problem when we
+need to know the return type of `f`, but must not determine this type by
+any actual evaluation of `f`. Because this problem does not occur in the
+non-null case, it does arise for either `Union{T, Void}` or
```
tkelman, line 330: it does arise or does not?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+much verbosity.
+
+### `Union{T, Null{T}}` Pros & Cons
+
+This final approach allows us to distinguish `Null{Bool}` from `Null{Int}` in
+cases in which that distinction might be useful. R makes such a distinction,
+although it seems to be done only because it is falls out naturally from an
+implementation that uses distinct sentinel values for each nullable type in
+the language. Otherwise, this approach is similar to the `Union{T, Void}` case
+described above, except for one crucial problem that would seem to suggest this
```
andyferris, line 285: For balance it would be fair to also state it's one "crucial" advantage:

`Union{T1, Null{T1}}` and `Union{T2, Null{T2}}` are disjoint, removing one major class of ambiguity problem for `f(x::?T1)` and `f(x::?T2)` which exists for the other two options.

(actually I just realized that it could be even better if `Null{T}` could be covariant in `T` - that also helps in the case that inference for `return_type` is too wide... speculatively - maybe covariance is helpful for `Some{T}` too?).

A final 4th class of nullable would be `Union{Some{T}, Null{T}}` - has that been discussed?

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+
+Some such functions are:
+
+* `isnull(x::?T)`
+* `get(x::?T)`
+* `get(x::?T, default::T)`
+
+Because we would use broadcasting, lifting would happen via dot-syntax:
+for example, `x .+ y` would be lifted by default. In addition, we would special
+several operators like `+`, `-`, `==`, etc. to automatically be equivalent to
```
tkelman, line 377: `==` gets into the 3VL discussion which isn't really touched on in the rest of this document, right?

tpapp:
Having run into the "counterfactual return type problem" a few times, I would very much like having a `Union{Some{T}, Void}`. However, I am wondering if the language features (with the exception of syntactic sugar) and compiler optimizations would generalize so that users would be able to efficiently implement similar constructs and related machinery (lifting etc). To make things concrete, I am thinking of types like `Union{Some{T},Err{E}}` following [result types in Rust](https://doc.rust-lang.org/std/result/). Note that I am not suggesting that `Base` does this, just that libraries can experiment with such constructs and get efficient code.

andyferris:
Yes, the plan would be to make all `Union{A,B}` types as fast as they can be as a generic compiler optimization, so packages should be able to experiment with whatever return types they prefer (of course, it would be nice if the ecosystem would play nicely together).

datnamer:
Do you think this would happen pre 1.0?

johnmyleswhite:
If this doesn't happen 1.0 it's hard for me to imagine it changing later on since we'll get stuck with a lot of design decisions. But it's ultimately @vtjnash's call since he has the expertise to make this happen that I lack.

vtjnash:
I'm hoping to (and working on) getting thr start of this into v0.6

Nullable.md
```diff
@@ -0,0 +1,442 @@
...
+non-nullable types. When `T` is `Int`, this path is any function on integers.
+Many such functions exist and it is possible that users will assume that
+`f(x::Void)` will also exist because `f(x::Int)` exists.
+
+In DataArrays, the closely related problem of needing definitions of
+`f(x::NAtype) = NA` for all functions `f` was addressed by manually defining
+`f(x::NAtype) = NA` for all functions `f` that commonly occurred in practice.
+Although this piece-meal approach worked effectively in many cases because the
+majority of real code makes use of only a small number of core functions,
+there was always yet another function that to had be defined manually when a
```
shashi, line 141: "had to be"

TotalVerb:
@nalimilan mentioned in #20502 that `Union{T, Null}`, with special support so that `f(::Null) = Null()` is automatically defined for every function, may help resolve the difficulties of #20502.

This is not currently discussed here (though possibly it has been discussed in side channels), but is definitely worth consideration.

johnmyleswhite:
Can we get input from @JeffBezanson on whether or not he's willing to allow a universal function definition of `f(x::Null) = Null()` for all `f`?

quinnj:
I'll add in a quick 2 cents: a while back, @StefanKarpinski and I were discussing this a little and kind of came to the conclusion that this should be ok, since we're basically introducing an _additional_ form of missingness/nullability into the language, apart from `nothing::Void`. In that way, `nothing` continues to be `nothing`, no behavior changed and it remains the "language-level" form of nothingness, with very little behavior defined on it.

`null::Null`, however, would be come a more substantial form of missingness with more behaviors defined.

Not sure if that actually helps the discussion any, but it seemed to clear things up in my mind in that we're not trying to mess w/ `nothing` and how it currently works in the language, just introduce an additional form of nothingness w/ more behavior.

TotalVerb:
Let me add something to ensure that all potential questions have answers.

One difficulty with that approach is distinguishing when an `f(x) = ...` definition intends to lift over `x::Null` or otherwise. For instance, we might want `string(Null()) == "Null()"`. Now if someone defines

```
foo(x) = length(string(x))
```

is it clear whether this should be `Null()` or `6` for `foo(Null())`?

TotalVerb:
Unrelated to the current matter of discussion; but one thing we should do in 0.6 if at all still possible, is deprecate using the ternary operator without spaces. Otherwise things like

```
@foo x ?Int :sym
```

will parse in an unexpected way.

TotalVerb:
There are also other functions which we must consider `Null` to not lift over. I wonder if maintaining such a list of functions might be just as much work as maintaining a list of functions it should lift over.

- `string`, `print`, `println`, `show`, `dump`, `display` etc.; all IO functions
- by extension, functions that call `string`, e.g. `lpad`. annoyingly, we might want to lift `lpad` over `Null()` as the first argument
- `Base.vect`, i.e. `[x, y]`; also `vcat` `hcat` `hvcat` `cat`
- `push!` `unshift!` `setindex!` `fill!` etc. in fact basically every function with a bang
- `in` `contains`
- `=>` e.g. for `Dict(x => Null())`, `Set`
- `isnull`, `get`, `filter`, `map`
- most (but not all? e.g. `Complex`) type constructors
- `ifelse`, `isa`, `typeof`, `Ref`, and most other builtins

And then of course a list of functions where the lifting behaviour is ambiguous

- `searchsorted` and friends
- `|>`, `==`, `!=`, `isequal`
- `run`, `close`, `sleep`, `wait` and other "non-bang but still imperative" functions for which a null argument doesn't seem to make sense

I am sure this list is incomplete; but it must be something to consider.

ararslan:
The more I think about it, the more I think we shouldn't lift by default; cases where `f(Null)` should _not_ be `Null` seem to be equally prevalent if not more so than those where it should be `Null`. (Plus, I don't think Jeff has changed his stance on automatic lifting for Base functions.) It's easy enough to determine the subset of Base functions which should provide lifting, and packages can opt to provide functionality for their functions as well. But making a broad assumption on what `f(Null)` should be seems a little dangerous.

TotalVerb:
There seems to be some consensus recently of `Nullable` going back to what it used to be, with `NAType`, except now with compiler support so it's faster. I wonder if the name should also change to `NAType`, as this seems in some ways to be moving away from what is called "null" in other programming languages.

ararslan:
I don't think it matters too much how we spell it so long as the behavior is well defined and well documented.

andyferris:
I also think "universal" default definitions (lifting) for `Null` would be undesirable. I think there are many situations where an error is better than continuing to propagate the `Null`.

I still believe that `Null{T}` would be a nicer semantic. Some functions support a wide range of signatures and not having the `T` means it's ambiguous which method to correspond with (and there may exist important semantic differences between the different method signatures). I feel that would be a real problem and become a major pain point when you are doing anything more complex than basic operations like `+`. The other nice thing about this that you can (safely) define operations on `Null{T}` such as `&(x::Bool, ::Null{Bool}) = x ? Null{Bool}() : false` for e.g. 3 valued logic (and users can do the same *safely* for a variety of types). Without the `T` then we should assume the `Null` value might be of `Any` type, so I wouldn't like to see 3-valued logic implemented with `&(::Bool, ::Null)` since users are free to overload the behavior of `&(::Bool, ::T)` for their own types `T`.

However, to get this *semantic* of a missing value of type `T`, we might benefit from some compiler support (specifically, from the type system). Ideally, `Null{T}` would be covariant in `T` to make working with non-concrete types easier and more consistent. For example, if `Null{Int}` represents some missing value of type `Int`, then I wonder if this missing value should also be considered of type `Null{Integer}`? That is, given that types represent sets of possible values, for every (concrete?) type `T` we create the nullable set by adding an extra member of the set `Null{T}()` which represents a missing value (let's name here the combined set as type `Nullable{T} = Union{T, Null{T}}`, though I think we should find a new name like `?T` or something). We could imagine the set `Nullable{S}` (where `S` is an abstract supertype of `T`) as including the value `Null{T}()` (as well as `Null{S}()`?). (~~Does this make any sense, @JeffBezanson?~~ no it does not)

If we *don't* get this behavior then a signature like `f(::Nullable{Integer})` would fail for `Null{Int}()` and I do not believe that accurately captures the intentions of the author of `f`. If we did implement this covariant behavior, we *do* get that `Null` (being `Null{T} where T`) and `Null{Any}` are more-or-less equivalent (as may be `Any` and `Nullable{Any}`...), so in many cases `::Null` would still be useful to put in signatures and there shouldn't be undue burden on users/package authors unless they actually *require* the type specificity to make dispatch consistent.

EDIT: What I said here about covariance of `Null{T}` is rubbish. `Null{<:T}` seems to have all the desired properties, e.g. `Null{<:Integer}` includes `Null{Int64}()`, `Null{Unsigned}()`, `Null{Integer}()`, `Null{Union{UInt8, Int8}}()`, etc.

ararslan:
@andyferris I don't see how `Null{T}` would differ significantly, or in any beneficial way, from the current `Nullable{T}`.

> Some functions support a wide range of signatures and not having the T means it's ambiguous which method to correspond with

That's why you define proper promotions in Base and locally define `f(::Null)` to be whatever makes sense. If not knowing the type when calling a method on a null value is a problem, I think that's more indicative of a design flaw.

andyferris:
> If not knowing the type when calling a method on a null value is a problem, I think that's more indicative of a design flaw.

Sorry, some functions are overloaded with lots and lots of methods on purpose, e.g. try `methods(+)`. Making use of multiple dispatch is the whole point of Julia, and null values of certain types will follow distinctive propagation rules. If we allow for e.g. 3-valued logic,
```julia
&(x::Bool, ::Null) = x ? Null() : false
&(::Null, y::Bool) = y ? Null() : false
&(::Null, ::Null) = Null()
```
and another user creates a type `T` which 

 * overloads `&` 
 * interacts with `Bool`, such as `&(::Bool, ::T) -> T`
 * they want to use `Union{T,Null}` and propagate it through `&(::Union{Bool,Null}, ::Union{T,Null})`

then they will *not ever* be able to stop `&(false, Null)` from being `false::Bool` instead of some value of `T` or `Null`. 

I would prefer we didn't add things which constrain multiple dispatch and user's ability to overload methods, whenever they decide they also need to use missing values. In the example above, the user would need to define another `&`-like *function* instead of adding a *method* to `&`. It seems, to me, to  be against the Julian way.

I do agree that for typical, database stype operations, you're not going to run into this esoteric set of conditions. But the fact is that we can foresee circumstances where using the type and dispatch system more strongly will lead to bad interactions with `Null`.

andyferris:
In short terms, if a user defines `T` such that `&(::Bool, ::T) -> T`, they might expect to be able to enforce that `&(::?Bool, ::?T) -> ?T`. However, if anyone implements 3-valued logic on `Bool`, *they won't be allowed to do that*.

ararslan:
That doesn't seem like a problem to me; that's just kind of how I expect `Union`s to work. The behavior is clear when you think in terms of the methods defined for the given function. We just have to make sure the behavior is well documented, then people will know they can't expect to be able to enforce `Union{Bool, Null} & Union{T, Null} = Union{T, Null}`.

andyferris:
> That doesn't seem like a problem to me; that's just kind of how I expect `Union`s to work. 

Of course, this is *exactly* how `Union{T,Null}` would work. I was just suggesting that `Union{T,Null{T}}` would follow the same `Union` rules but is more flexible for users.

Do note that the covariance idea means that 99.9% of the time you can just use `Null` and `?T` for dispatch, and any users that want `Union{Bool, Null{Bool}} & Union{T, Null{T}} = Union{T, Null{T}}` can still get that. It just seems like a more flexible solution, is all. The other problem with untyped `Null` and `?T = Union{T,Null}` is that dispatch would have many ambiguities if users use `::?T` in signatures. I can imagine that I'd like to write:
```julia
f(x::?T1) = if isnull(x) ? #= do a =# : #= do b =#  # note that branch is elided by compiler!
f(x::?T2) = if isnull(x) ? #= do c =# : #= do d =#
f(Null) # ambiguous
```
That basically means it's *never* safe to dispatch on `?T` since two users/packages (who might not be collaborating) might both overload `Base.f(::?T)` for same `f` and different `T`, and end up with this ambiguity problem. This seems (to me at least) as a major flaw.

>I don't see how `Null{T}` would differ significantly, or in any beneficial way, from the current `Nullable{T}`.

I'm arguing that there is (or that there might be) a benefit for `Null{T}` over `Null`. I'm wondering what people see as the downsides (especially w.r.t. dispatching on `?T`)?

EDIT: please replace "covariance idea" with "using `Null{<:T}`" when reading the above.

andyferris:
> > I don't see how `Null{T}` would differ significantly, or in any beneficial way, from the current `Nullable{T}`.

> I'm arguing that there is (or that there might be) a benefit for `Null{T}` over `Null`. I'm wondering what people see as the downsides (especially w.r.t. dispatching on `?T`)?

Or to answer your question another way, the only problem I saw with `Nullable` was that it is a container. `Null{T}()` is not a container, it's a value, just as `Null()` would be. 

What are the problems that `Null{T}()` inherits from `Nullable{T}()`? Is it that you might need to sometimes use `promote_op` and so-on? (I actually think you won't need that nearly as much for `Null{T}` as you would for `Nullable{T}`, since you can just treat it as a value. Only containers of `?T` would use `promote_op` just as they do for `T` itself).

johnmyleswhite:
The relevant problem with `Nullable{T}` is also the problem with `Null{T}`: they're parametric types. Being a container or a union is not the problem: the problem is computing the type parameters of a parametric type in the absence of type inference.

andyferris:
> The relevant problem with `Nullable{T}` is also the problem with `Null{T}`: they're parametric types. Being a container or a union is not the problem: the problem is computing the type parameters of a parametric type in the absence of type inference.

Actually it's only parametric *containers* that rely on inference.

You can for instance define `+(::Null{Int32}, ::Null{Int64}) = Null{Int64}()`. No inference required. Just dispatch.

You *could* use `promote` or `promote_op(+, ...)` to have a generic method for `+(::Null{T1}, Null{T2})`, but unlike the container situation, you are not forced to, and I would suggest we don't use `promote_op` (`promote` might be OK, wherever it is already used?).

johnmyleswhite:
I give up. I'm done having these debates when no one involved is paying me for my time.

andyferris:
John - please don't take me the wrong way. We *all* really appreciate the effort that is going into this - I think it is really fantastic!

(I'm only trying to bounce some ideas around and have a technical discussion about some things which I thought might have been overlooked)

TotalVerb:
@johnmyleswhite I too greatly appreciate the effort that has gone into this Julep. Thanks for all your hard work; it's truly important for building a strong foundation for 1.0.

@andyferris There is a flaw, in my opinion, with the covariant `Null{T}` idea. It makes the assumption that every _type_ on its own represents a coherent algebraic structure, to which we can add a unique null type. By making `Null{T}()` an instance iff `T` is concrete, we lose the ability to adjoin a null element to algebraic structures which are, for whatever reason, represented by an abstract type.

Consider for example

```julia
abstract type Symbolic <: Real end
struct SymInteger <: Symbolic; val::BigInt; end
struct Reciprocal <: Symbolic; of::Symbolic; end
```

and imagine that, conventionally, `Symbolic` is used to denote some real number represented symbolically in this fashion. Then, what we desire is a `Null{Symbolic}()` constant. But with `Null{T}()` existing iff `T` concrete, what we instead have are two disjoint `Null{SymInteger}()` and `Null{Reciprocal}()` types, which does not seem fun to have to deal with.

andyferris:
@TotalVerb I think you might be right (I think there was a ? in my post next to the "concrete"). Can we have consistent covariance (`Null{SymInteger}()::Null{Symbolic}`) as well as a `Null{Symbolic}()` value?

TotalVerb:
Hmm, that would be unusual compared to anything else in the language (it would be a type which has an instance, but also has subtypes), but I suppose there is nothing preventing its implementation. Moving on to practical issues though, as mentioned in the Julep, consider

```julia
function map(f, x::Nullable{T})
    if !isnull(x)
        return Nullable(f(get(x)))
    else
        return Nullable{Core.Inference.return_type(f, (T, ))}()
    end
end
```

If we rewrite this with `Union{T, Null{T}}` we obtain

```julia
map(f, x::Any) = f(x)
map{T}(f, x::Null{T}) = Null{Core.Inference.return_type(f, Tuple{T})}()
```

which not only has an inference dependency, but also has a strange definition of map. It seems like we would be giving up on `map` entirely with `Union{T, Null{T}}`. This would be  a departure from [Scala](http://www.scala-lang.org/files/archive/nightly/docs/library/index.html#scala.Option), [Java](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html), [Rust](https://doc.rust-lang.org/std/option/enum.Option.html#method.map), [F#](http://fsharpforfunandprofit.com/posts/the-option-type/), [Swift](http://www.javiersoto.me/post/106875422394), [Kotlin](https://yobriefca.se/blog/2012/07/17/kotlin-nullable-extensions/), {Ca|Standard}ML. While I'm not saying that is necessarily a bad thing, we should consider why `map` is considered useful in such languages, and see if we have a replacement for those use cases.

TotalVerb:
An instructive exercise, perhaps, is to find some concrete examples of code that uses `Nullable`s in practice, and consider how it would be written using each suggested interpretation. The applications in `NullableArrays` are well-known and of course well-studied, but we should consider other applications also.

Let me get the ball rolling with some parts of Base Julia that use `Nullable` or an equivalent type.

- `match` in Base returns `nothing` when there is no match. (This is effectively `Union{T, Void}` or `Union{T, Null}`.) As far as I can tell, any code using conditional regex matches is practically forced to use an if statement, because there is neither autolifting nor opt-in lifting for the `Void` in `Base.
- `get(dict::Associative, key)` as proposed in https://github.com/JuliaLang/julia/pull/18211. In particular, this is not only a convenience issue, but also a performance one, because this can avoid the double hash lookup.
- [The revised iteration protocol](https://github.com/JuliaLang/julia/issues/18823) gives yet another perspective.

I suspect that a lot of these use cases will be relatively independent of which interpretation is chosen, but we should also consider package code that makes use of nullable.

andyferris:
> Hmm, that would be unusual compared to anything else in the language (it would be a type which has an instance, but also has subtypes), but I suppose there is nothing preventing its implementation.

Actually that does seem like a really big departure from the current type system.

> If we rewrite this (`map`) with Union{T, Null{T}} we obtain

Do people need `map(f, x::Union{T, Null})` for values? If `T` is not a container; you just call `f(x)`. Of course, you can do `map(sin, 1)` and for the null case we can make `Null` an iterable of length 1, just like `Int`. 

If `T` is a container such as `T <: AbstractArray`, say, then we'd need an extra method for `map` for that, and it might invoke inference on the element types if that is what it does now for empty arrays anyway... seems equivalent.

TotalVerb:
I think you're right. `map` is for containers; `Union{T, Null{T}}` is not a container and so makes it not generally necessary.

However, this goes back to the original issue of when functions should be lifted over null arguments. If I want to compute `f.(x::Nullable{T})`, whether or not the person who defined function `f` defined `f(::Null{T})`, there should still be a way to do that; otherwise we are losing the functionality of `map`. This is, modulo the inference dependency, I think fairly easy to resolve:

```julia
lift(f, x) = f(x)
lift(f, x::Null{T}) = Null{C.I.return_type(f, Tuple{T})}()
```

However, because there is no more `Some{T}` tag indicating that something was part of a `Nullable`, this means that we need to use a new syntax, and can no longer reuse the `broadcast` dot syntax.

andyferris:
Thank you very much @TotalVerb for helping me figure out the trade-offs here. It seems true that the `Null{T}` approach does seem to preclude "lifting" methods (or entire functions) without performing a type inference step.

My general approach here has been to see if sufficiently clever dispatch would mean lifting is not necessary - and demanding that users/package authors provide the necessary method. I will concede that this might not be very user-friendly, especially in the case that the methods are coming from somebody else's package. In any case, I would suggest providing a range of common methods for `Null` or `Null{T}` so that lifting is relatively rare. And, if lifting is relatively rare it (speculatively) seems not too ridiculous to invoke inference in these cases? I dunno.

Anyway, it seems that `Null` is better suited for an explicitly annotated lifting approach and `Null{T}` is more geared towards dispatch.

andyferris:
By the way, everything I was saying about the covariance of `Null{T}` is rubbish. In Julia v0.6 it's dead easy to express `Null{<:Integer}` which includes `Null{Int64}()`, `Null{Unsigned}()`, `Null{Integer}()`, `Null{Union{UInt8, Int8}}()`, etc.

Thus I would suggest having `?T = Union{T, Null{<:T}}` (or whatever the syntax for the left-hand side here becomes). Then dispatch for `f(::?Integer)` would be *quite* sensible. No special compiler/type-system treatment required.

nalimilan:
> My general approach here has been to see if sufficiently clever dispatch would mean lifting is not necessary - and demanding that users/package authors provide the necessary method. I will concede that this might not be very user-friendly, especially in the case that the methods are coming from somebody else's package. In any case, I would suggest providing a range of common methods for Null or Null{T} so that lifting is relatively rare. And, if lifting is relatively rare it (speculatively) seems not too ridiculous to invoke inference in these cases? I dunno.

 @andyferris This approach has already been tried with vectorized methods, and it was considered as a design flaw. In 0.6 we've been able to move away from it by using the `f.(x)` syntax for element-wise operation. So we've followed that route already and we know it won't work. If a user needs to pass a null values to a function provided by a package which doesn't support it, s/he shouldn't need to rewrite the function. I think repeating these discussions once again is one of the reasons John closed the PR.

Also the issue with inference is a deep one. Even if we get rid of `map` support for nullables, lifting will have by definition to infer the `T` for `Null{T}` if we make it parametric. It's not OK for it to fail in some cases, as these cases will behave differently from others if you dispatch on the inferred type. What's the point of opting for `Null{T}` if we can't rely on it? Note that the case of empty collections is considered as fragile, but we haven't found a better solution. Making it the foundation of nullables isn't the best idea to write a robust system.

> That basically means it's never safe to dispatch on ?T since two users/packages (who might not be collaborating) might both overload Base.f(::?T) for same f and different T, and end up with this ambiguity problem. This seems (to me at least) as a major flaw.

Yes, and that shows the only solution in that case is to have a common definition `f(::Null) = Null()`. That's why I suggest we do this by default for all functions.

(The example of 3VL `&` and `|` is not representative at all, since these are the only two functions I know for which we want lifting, but not with the standard lifting semantics. So let's not base the whole system on this special case, for which it could be OK to define methods manually.)


Finally, I'm not sure we need so many functions *not* to be lifted. Functions for which passing a nullable makes no sense should generally not be lifted, and people shouldn't pass nullables to them. Why would one call `lpad(Null())` or `run(Null())` and expect not getting `Null()` as a result?

Anyway, we could start without a default lifting fallback and see how it goes. At least, `Union{T, Null}` would be as functional as DataArrays or `Nullable`, but more convenient than the latter since non-null values would not even need unwrapping.

andyferris:
Thank you so much @nalimilan for your wonderful and very detailed response. I have certainly learned a lot today from this discussion. I had been interested in the development of `Nullable` and for quite some time have been following the public discussions, issues and this PR as best I can, yet I always was left with (central) one question: "why use lifting with `Union`-based nullables?" (as opposed to dispatch, which does require extra method definitions and certainly inconvenience, but nothing worse than e.g. C++ would be). I am very glad for receiving a full and direct explanation.

In any case, I feel that having this discussion has been valuable exercise. This certainly is a tricky design problem, and there are clearly rather few available options which satisfy e.g. determinism w.r.t. inference and user-friendliness.

@johnmyleswhite I'd like to apologize to you directly - as I alluded to earlier, I have been very impressed with the leadership shown with this PR and, in fact, have been looking forward to seeing this PR merge. It was never my intention to derail this process. However, I think it's reasonable to say this Julep has become a focus point for the future of `Nullable` and I assumed it would be an OK place to check my understanding and assumptions about this design, clear up some misunderstandings, float ideas, and in general, to come to a consensus. Thank you for your efforts - and I still believe this Julep process has been, and will continue to be, very worthwhile.

nalimilan:
I think we should merge this PR and add the missing details later.

PR #34, opened 2017-05-18
Remove bullet mentioning removal of findin

andreasnoack:
Algorithmic complexity considerations are important for the `find` family
of functions in general but for this function in particular. The complexity
of the current implementation `findin` O(m + n) on average with a fairly
large constant. The suggested alternative would have a complexity of O(m*n).

### DIFF
--- a/Find.md
+++ b/Find.md
@@ -173,9 +173,6 @@ exist in the current API.
 Other issues are more localized and can be fixed one by one, depending on the chosen general
 plan.
 
-- **`findin`**: This function is just a shorthand for `find(x -> x in c, A)`. It may not be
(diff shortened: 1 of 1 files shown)


ararslan (approved):

nalimilan (approved):
Good catch!

nalimilan:
Coming to this again, it looks like the `find(occursin(x), y)` solution used at https://github.com/JuliaLang/julia/pull/24673 could allow replacing `findin` with a specialized `find(::OccursIn, ::Any)` method, and still use the most efficient implementation under the hood (it will differ e.g. between arrays and strings).

Cc: @JeffBezanson

PR #40, opened 2017-06-20
WIP Binaries API Julep

iamed2:
TODO:
- [ ] Failures with HTTP error codes
- [ ] Deprecation/transition strategy/timeline
- [ ] Infrastructure proposal example
- [ ] Scope box

### DIFF
--- a/BinariesAPI.md
+++ b/BinariesAPI.md
@@ -1,0 +1,107 @@
+# JULEP BinariesAPI
+
+- **Title:** A structured HTTP API for getting Julia binaries
+- **Authors:** Eric Davies <<iamed2@gmail.com>>
+- **Created:** June 16, 2017
+- **Status:** proposed
+
+## Abstract
+
+The URLs for fetching Julia binaries of various versions have changed a few times across
+releases.
+The latest change necessitated PRs to multiple repos to change AppVeyor scripts.
+This Julep proposes a level of abstraction over the Julia binary URLs which is decoupled
+from the storage service used for the binaries.
+
+## Current Status
+
+The current (as of this Julep's creation) [posted URLs](https://julialang.org/downloads)
+for Julia binaries are (using 64-bit Linux and 32-bit Windows for example):
+
+- 0.5 (release)
+  - https://julialang-s3.julialang.org/bin/linux/x64/0.5/julia-0.5.2-linux-x86_64.tar.gz
+  - https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.2-win32.exe
+- 0.6 (release candidate)
+  - https://julialang-s3.julialang.org/bin/linux/x64/0.6/julia-0.6-latest-linux-x86_64.tar.gz
+  - https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe
+- 0.7 (nightly)
+  - https://status.julialang.org/download/linux-x86_64
+  - https://status.julialang.org/download/win32
+
+Recently, PRs were made to update binary URLs for AppVeyor scripts.
+The URL for the nightly build was different than the above:
+
+- https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe
+
+Source code links are also provided, linking to GitHub's auto-generated tarballs.
+
+## General Proposal
+
+Set up a domain (e.g., binaries.julialang.org) for hosting binaries, with the URLs isolated
+from the service hosting the files (e.g., Amazon API Gateway over S3).
+Introduce a standard format for the Julia binary URLs, respected by every release.
+
+## URL Format Proposal
+
+`https://binaries.julialang.org/<os>/<arch>/<version>`
+
+### OS
+
+`<os>` is one of `linux`, `windows`, `macos`, and any others which may be added in the
+future (e.g. `freebsd`).
+
+### Arch
+
+`<arch>` is one of `x86`, `x86_64`, `arm`, `ppc64le`, and any others which may be added in
+the future (e.g. `aarch64`).
+These are the abbreviations used in the
+[LLVM source](http://llvm.org/docs/doxygen/html/Triple_8h_source.html).
+
+### Version
+
+`<version>` is a Julia version prefix.
+Any trailing unspecified components are assumed to be the maximum available value once
+all other conditions are satisfied.
+
+#### Valid Version Values
+
+All values must be `release` or `nightly` or match this specification:
+
+```regex
+^v\d+(\.\d+)?(\.\d+)?(-\w+)?([\+\.]\w+)?$
+```
+
+Assuming the latest nightly is `v0.7.0-DEV.568`, the latest release is `v0.5.2`, and
+the latest 0.6 release candidate is `v0.6.0-rc3`:
+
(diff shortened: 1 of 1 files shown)


BinariesAPI.md
```diff
@@ -0,0 +1,107 @@
...
+- **Created:** June 16, 2017
+- **Status:** proposed
+
+## Abstract
+
+The URLs for fetching Julia binaries of various versions have changed a few times across
+releases.
+The latest change necessitated PRs to multiple repos to change AppVeyor scripts.
+This Julep proposes a level of abstraction over the Julia binary URLs which is decoupled
+from the storage service used for the binaries.
```
tkelman, line 14: the fastly caching is basically this already

BinariesAPI.md
```diff
@@ -0,0 +1,107 @@
...
+```
+
+Assuming the latest nightly is `v0.7.0-DEV.568`, the latest release is `v0.5.2`, and
+the latest 0.6 release candidate is `v0.6.0-rc3`:
+
+- `v0`, `v0.7`, `v0.7.0`, `v0.7.0-DEV`, and `v0.7.0-DEV.568` would match `v0.7.0-DEV.568`
+- `v0.6`, `v0.6.0`, and `v0.6.0-rc3` would match `v0.6.0-rc3`
+- `v0.5` and `v0.5.2` would match `v0.5.2`
+- `v0.5.0` would match `v0.5.0`
+- `release` would match `v0.5.2`
```
tkelman, line 81: this is redundant (or a bad idea to use) if we also have a single-digit major link

tkelman:
See https://github.com/JuliaLang/julia/pull/18554

iamed2:
Oops, I did not intend to open this already. Well, oh well.

BinariesAPI.md
iamed2: @omus and I were talking about this. 

Suppose a user created a package in a completed state and wanted to ensure it would remain compatible with the latest release for eternity. They could enable release and nightly builds using Travis' cron feature, with the nightly builds set to allow failures but notify that user. They could fix any issue with the nightly build if it occurred and be assured that the package would be release-compatible.

BinariesAPI.md
iamed2: That's not a thing I know about and I cannot find any information on it.

BinariesAPI.md
tkelman: Travis settings aren't what determines the supported versions of Julia though. Is there realistically going to be a package that never gets touched for longer than the lifetime of multiple major versions of Julia but will need to be automatically tested against major versions that don't exist yet?

BinariesAPI.md
iamed2: But not quite. There are still two domains at play. And AFAICT it still relies on a file structure (with one copy of each file added for each alias).

rofinn: @tkelman are you just referring to the switch from `s3.amazon.com/julialang/` to `julialang-s3.julialang.org`? I think the proposal is just for that url schema to stabilize and be consistent with nightlies?

tkelman: What's the value in adding a third? If we're going to pick a url structure and naming convention that we intend to keep stable forever, the caching layer can be laid out that way.

BinariesAPI.md
```diff
@@ -0,0 +1,107 @@
+# JULEP BinariesAPI
+
+- **Title:** A structured HTTP API for getting Julia binaries
+- **Authors:** Eric Davies <<iamed2@gmail.com>>
+- **Created:** June 16, 2017
+- **Status:** proposed
+
+## Abstract
+
+The URLs for fetching Julia binaries of various versions have changed a few times across
```
rofinn, line 10: I'll note that I've needed to update Playground.jl a few time for this. Not a huge issue, but it would be nice if there was a stable url schema.

BinariesAPI.md
iamed2: If the Travis settings corresponded to these alias URLs, wouldn't they be then coupled to the supported versions of Julia?

> Is there realistically going to be a package that never gets touched for longer than the lifetime of multiple major versions of Julia but will need to be automatically tested against major versions that don't exist yet?

I'm not sure what you mean by "major versions that don't exist yet". But I believe https://github.com/andrewcooke/AutoHashEquals.jl and https://github.com/iamed2/ResultTypes.jl are packages like that.

BinariesAPI.md
iamed2: Also, any change to the storage would require a change to the URLs, so there is still coupling in that sense.

BinariesAPI.md
tkelman: When has the S3 layout ever changed? It's been constant since at least 0.2.

BinariesAPI.md
rofinn: I agree, but I think the proposal could be interpreted as making the conventions in the caching layer stable (and supporting nightlies with the same schema).

BinariesAPI.md
tkelman: Pkg determines supported versions of Julia from REQUIRE, a separate file from .travis.yml. A `release` key would stop testing against 1.x.y as soon as 2.0.0 gets released. This is the reason I've systematically opened hundreds of PR's eliminating usage of the `release` key on Travis, and I intend to remove it completely.

iamed2: Ah I misunderstood what you meant by "supported versions". 

A possible configuration:

`REQUIRE` has a minimum Julia version of 1.0.0

`.travis.yml` tests on 1.0.0 to ensure minimum version compatibility, and then also `release` and `nightly` to match new-user expectations (new users will expect packages to work on the most recent version on the website).

tkelman: `release` would end up frequently redundant. This might be better served by revisiting PkgEval automatic issue filing instead of adding to the matrix that runs on every single commit on Travis. In terms of user expectation, the latest tagged release is the more relevant thing to want to run release-julia tests against on a cron schedule, not master of the package.

BinariesAPI.md
rofinn: Well I recall the use of `pre` vs `latest` was used for 0.4 pre-releases. I'll admit that some of the work on my part was due to lack of knowledge/documentation (I recall parsing the downloads page at one point).

BinariesAPI.md
iamed2: > release would end up frequently redundant

Wouldn't it never be redundant once `1.0.1` was released?

> adding to the matrix that runs on every single commit on Travis

It's possible to enable certain builds for only tags, PRs, or a certain branch, for example. I am not arguing to keep `release` included by default in `.travis.yml` in generated packages.

BinariesAPI.md
tkelman: the exact link used on the downloads page has changed a few times, and the 0.x-latest copies did not always exist

BinariesAPI.md
tkelman: It will be redundant except in the rare occasion that a package happens to hit a bug that gets fixed in between 1.0.0 and 1.0.1 (or starts relying on a feature that is in 1.1.0 but not 1.0.0), in which case it would be good to know when it happens, but this is more of a release concern than an every-commit concern.

PR #45, opened 2017-09-01
add another proposal (`findeach`) to the find julep

JeffBezanson:
From https://github.com/JuliaLang/julia/issues/10593#issuecomment-326347628

### DIFF
--- a/Find.md
+++ b/Find.md
@@ -168,6 +168,36 @@ droppped in favor of iterators.
 \* These combinations are not needed as they correspond to `searchseq`. Indeed they do not
 exist in the current API.
 
+## Proposal 3
+
+This proposal adds `findeach(pred, A[, rev])`, which returns an iterator and can be used to
+implement most of the other functions in one line.
+Predicates are always used instead of separate functions for different kinds of searches when possible.
+This potentially allows using the same function for sequence searching, since a subsequence to look for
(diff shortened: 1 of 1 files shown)


Find.md
```diff
@@ -168,6 +168,36 @@ droppped in favor of iterators.
...
+the general iterator needs to operate on state objects.
+
+We can also keep `findfirst` and `findnext`, since they are especially convenient.
+Ideally we will keep only `findfirst(pred, A)` and deprecate other methods.
+
+If we want, `find` can be deprecated to `collect(findeach(...))`.
+
+The following functions can also be deprecated to `findeach` calls: `findin`, `search`, `rsearch`, `match`, `eachmatch`.
+
+This proposal does not touch `findmin`, `findmax`, etc.
```
nalimilan, line 199: This is redundant with the general comment in the next section: no proposal touches these functions.

nalimilan (commented):
Thanks! I like this radical proposal which simplifies things a lot.

The only point I find problematic is searching for a sequence: contrary to regexes, sequences don't have a dedicated type, so ambiguities could arise (or at least confusion in users' minds). So I'd rather use a special predicate for that. Anyway, since these methods do not exist in the current API (only for strings, which is less of a problem) we can always tackle this later.

PR #50, opened 2018-06-28
Julep to enhance GC interface for foreign code.

rbehrends:
The support for modules written entirely or partly in foreign languages
to interface with the Julia GC or to use the GC for allocations that do
not neatly fit Julia's type system or use low-level approaches not
available in Julia (such as irregular data structure layouts) is
currently somewhat limited.

This proposal aims at fleshing out the API for allowing more complex
interaction of foreign code with the GC, especially the use of long-lived
foreign objects that are inextricably interwoven with Julia
objects.

### DIFF
--- a/GcExtensions.md
+++ b/GcExtensions.md
@@ -20,19 +20,19 @@ objects.
 
 Specific use cases that we are trying to address are:
 
-* *Allowing the Julia GC to manage foreign objects with arbitrary
+1. *Allowing the Julia GC to manage foreign objects with arbitrary
   layouts.* Not all objects -- especially those from preexisting
   libraries -- fit Julia's type system, for example, specialized
   container types written in C/C++. Such objects can comprise multiple
   memory blocks that require a custom marking mechanism and may also
-  require finalizers written in C.
-* *Providing additional roots to the GC.* Currently, to have additional
+  require low-level finalizer behavior written in C.
+2. *Providing additional roots to the GC.* Currently, to have additional
   roots, they must be stored in a location that is visible to Julia.
   This can be expensive if such roots are updated frequently or are
   contained in data structures that would have to be laboriously
   translated into a format usable by Julia. Instead, we want to allow
   for roots to be discoverable at the beginning of a garbage collection.
-* *Conservative scanning of stack frames and objects.* Currently,
+3. *Conservative scanning of stack frames and objects.* Currently,
   scanning does have to be precise. If we desire to use the GC for
   foreign code that requires conservative scanning (especially for
   foreign stack frames), then it is necessary to have functionality
@@ -54,13 +54,13 @@ discussion of overhead can be found accompanying the descriptions of
 these hooks.
 
 An implementation of this proposal can be found on GitHub under
-https://github.com/rbehrends/julia (branch `rb/gc-extensions`). See
+<https://github.com/rbehrends/julia> (branch `rb/gc-extensions`). See
 the example in the `test/gcext` subdirectory for an example of using
 this API.
 
 An implementation of GAP that uses the Julia GC in lieu of its native
-GC can likewise be found on GitHub at https://github.com/rbehrends/gap
-(branch `alt-gc`). This version of GAP can be built with:
+GC can likewise been found on GitHub at <https://github.com/rbehrends/gap>
+(branch `alt-gc`). That version of GAP can be built with:
 
     ./autogen.sh
     ./configure --with-gc=julia --with-julia=/path/to/julia/usr
@@ -72,39 +72,50 @@ In order to allow foreign code to have access to necessary functionality
 in the garbage collector, we allow foreign code to register callbacks for
 certain GC events. We provide for six types of callbacks:
 
-* `pre_gc` (beginning of garbage collection).
-* `post_gc` (end of garbage collection).
-* `root_scanner` (before scanning GC roots).
-* `task_scanner` (when scanning Julia tasks).
-* `notify_external_alloc` (when an external object is allocated).
-* `notify_external_free` (when an external object is freed).
+1. Beginning of garbage collection (`pre_gc`)
+2. End of garbage collection (`post_gc`)
+3. When scanning GC roots (`root_scanner`)
+4. When scanning Julia tasks (`task_scanner`)
+5. When an external object is allocated (`notify_external_alloc`).
+6. When an external object is deallocated (`notify_external_free`).
 
-With external objects, we refer to what in the current implementation are
+These callbacks are *not* per se thread-safe. It is up to to the callback
+implementation to ensure that no violations of thread-safety occur.
+
+In particular, each of these can be called from any thread. All except the
+first two can be called concurrently. In the current Julia GC implementation,
+the `post_gc` callback may also not be called before the next `pre_gc`.
+
+With external objects, we refer to what in the current Julia implementation are
 called `bigval_t` objects. These are allocated using the system's memory
 allocator rather than using Julia's external allocator. In order to not
-expose the implementation detail, we talk about "internal" and "external"
+expose this implementation detail, we talk about "internal" and "external"
 objects rather than objects that are allocated as part of Julia's object
 pool or through system routines, respectively.
 
 For each type of callback, there is a corresponding function pointer type.
-Registering and deregistering callbacks occurs via macros.
+Registering and deregistering callbacks occurs via corresponding setter
+functions.
 
 ```
 typedef void (*jl_gc_cb_pre_gc_t)(int full);
 typedef void (*jl_gc_cb_post_gc_t)(int full);
 typedef void (*jl_gc_cb_root_scanner_t)(int full);
 typedef void (*jl_gc_cb_task_scanner_t)(jl_task_t *task, int full);
-typedef void (*jl_gc_cb_external_alloc_t)(void *addr, size_t size);
-typedef void (*jl_gc_cb_external_free_t)(void *addr);
+typedef void (*jl_gc_cb_notify_external_alloc_t)(void *addr, size_t size);
+typedef void (*jl_gc_cb_notify_external_free_t)(void *addr);
 
-#define jl_gc_register_callback(cb, func)
-#define jl_gc_deregister_callback(cb, func)
+void jl_gc_set_cb_root_scanner(jl_gc_cb_root_scanner_t cb, int enable);
+void jl_gc_set_cb_task_scanner(jl_gc_cb_task_scanner_t cb, int enable);
+void jl_gc_set_cb_pre_gc(jl_gc_cb_pre_gc_t cb, int enable);
+void jl_gc_set_cb_post_gc(jl_gc_cb_post_gc_t cb, int enable);
+void jl_gc_set_cb_notify_external_alloc(jl_gc_cb_notify_external_alloc_t cb, int enable);
+void jl_gc_set_cb_notify_external_free(jl_gc_cb_notify_external_free_t cb, int enable);
 ```
 
-The `jl_gc_register_callback()` and `jl_gc_deregister_callback()` macros
-are called with the appropriate index and a function pointer of the
-matching type. Each function can be registered at most once; duplicate
-registrations will be discarded.
+For each setter function, a callback function is supplied, along with a flag
+(`1` for enabling the callback, `0` for removing it again). Attempting to
+register a callback multiple times will only register it once.
 
 *Performance impact:* The callback implementation is designed to incur
 negligible overhead if no callbacks are used and no more overhead than
@@ -117,16 +128,20 @@ long as the variable is in the cache and the branch target in the BTB.
 ## Additional GC roots and hooking into the GC process
 
 We provide three callbacks that are called at the beginning of a GC
-(`pre_gc`), the beginning of the mark phase (`root_scanner`),
-and the end of the GC (`post_gc`).
+(`pre_gc`), the beginning of the mark phase (`root_scanner`), and the end of
+the GC (`post_gc`). As these callbacks are tested and called only once per
+collection, overhead should be negligible. The `full` argument passed
+to these callbacks indicates whether this is a full or partial garbage
+collection.
 
-As these callbacks are tested and called only once per collection, overhead
-should be negligible.
+In addition, we also provide a `task_scanner` hook, which functions like
+the `root_scanner` hook, except that it is called for each task and with
+a pointer to the task object as its first argument.
 
-Additional roots can be marked from the root scanner callback by
-calling the `jl_gc_mark_queue_obj()` function, which takes a pointer to
-the current thread's thread-local storage a pointer to the object as its
-parameters.
+Additional roots can be marked from the `root_scanner` and
+`task_scanner` callbacks by calling the `jl_gc_mark_queue_obj()`
+function, which takes a pointer to the current thread's thread-local
+storage a pointer to the object as its parameters.
 
 ```
 int jl_gc_mark_queue_obj(jl_ptls_t ptls, jl_value_t *obj);
@@ -139,37 +154,66 @@ the thread-local storage of the current thread.
 The return value of `jl_gc_mark_queue_obj()` can be ignored for marking
 roots, but will be relevant for marking foreign objects (see below).
 
+When processing large objects, calling `jl_gc_mark_queue_obj()` can be
+ineffecient, as each object will be pushed on the mark stack separately.
+
+If possible, it is therefore recommended that programmers use the
+following function, designed for arrays of references, which handles
+this use case more efficiently:
+
+```
+void jl_gc_mark_queue_objarray(jl_ptls_t ptls, jl_value_t *parent,
+    jl_value_t **objs, size_t nobjs);
+```
+
+Here, `parent` is a reference to the current object, `objs` is a pointer
+to the start of an array of object references, and `nobjs` is the number
+of object references contained in that array. That array must be part of
+the object; it must not be allocated in static memory or on the stack.
+
+Unlike `jl_gc_mark_queue_obj()`, this function does not have a return
+value, as it does the requisite tracking itself.
+
+Calling this function will only require one slot on the mark stack, as
+opposed to the `nobjs` slot that individual calls to
+`jl_gc_mark_queue_obj()` would require, making it considerably more
+memory efficient.
+
 ## Managing foreign objects with custom layouts
 
 Foreign objects with custom layouts can define their own datatype through
-the `jl_new_foreign_type()` function. Its first three parameters are the
-same as for regular data types; following are a pointer to a mark function
-and a pointer to a finalizer function (the latter of which can be null).
-
-The mark function argument (`markfunc`) and the finalizer function
-(`finalizefunc`) take function pointer arguments; the finalizer function
-can be `NULL`.
-
-The `haspointers` parameter should be non-zero if
-the object may contain references to Julia objects; the `large` parameter
-should be non-zero if the size is greater than the value returned by
-`jl_gc_max_internal_obj_size()` and zero otherwise. If the object can be
-both larger or not, then two distinct foreign types need to be created, one
-for the case where the size is less than or equal and one for the case where
-it is larger than the value of `jl_gc_max_internal_obj_size()`.
+the `jl_new_foreign_type()` function:
 
 ```
 typedef uintptr_t (*jl_markfunc_t)(jl_ptls_t ptls, jl_value_t *obj);
-typedef void (*jl_finalizefunc_t)(jl_value_t *obj);
+typedef void (*jl_sweepfunc_t)(jl_value_t *obj);
+
 jl_datatype_t *jl_new_foreign_type(
   jl_sym_t *name,
   jl_module_t *module,
   jl_datatype_t *super,
   jl_markfunc_t markfunc,
-  jl_finalizefunc_t finalizefunc,
+  jl_sweepfunc_t sweepfunc,
   int haspointers,
   int large
 );
+```
+
+The first three parameters of `jl_new_foreign_type` are the same as for
+regular data types; following are a pointer to a mark function
+(`markfunc`) and a pointer to a sweep function (`sweepfunc`); the latter
+of which can be null.
+
+The `haspointers` parameter should be non-zero if instances of the new
+datatype may contain references to Julia objects; the `large` parameter
+should be non-zero if the size of instances of the new datatype will be
+greater than the value returned by `jl_gc_max_internal_obj_size()` and
+zero otherwise. If the objects can be both larger or not, then two
+distinct foreign types need to be created, one for the case where the
+size is less than or equal and one for the case where it is larger than
+the value of `jl_gc_max_internal_obj_size()`.
+
+```
 size_t jl_gc_max_internal_obj_size(void);
 ```
 
@@ -182,14 +226,17 @@ To accomplish that, such foreign types use the existing
 which is looked at after the other data types and the other alternatives
 for `fielddesc_type`.
 
-The mark function gets passed a pointer to thread-local storage (`ptls`)
+### Mark functions for foreign objects
+
+The mark function `markfunc` gets passed a pointer to thread-local
+storage (`ptls`)
 and the object to be marked (which will be of the type defined through
 `jl_new_foreign_type()`. The `ptls` argument is an optimization so that
 `jl_get_ptls_states()` does not need to be called unnecessarily during
 the mark loop.
 
 The mark function implementation also uses `jl_gc_mark_queue_obj()` to
-mark objects, as with the root scanner; however, in contrast to marking
+mark objects, as with the `root_scanner` callback; however, in contrast to marking
 roots, the return value cannot be ignored. Per object, the mark function
 should count how often `jl_gc_mark_queue_obj()` for subjects return
 non-zero values and return that number. If an object has no subobjects,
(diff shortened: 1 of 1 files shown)


GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+  This can be expensive if such roots are updated frequently or are
+  contained in data structures that would have to be laboriously
+  translated into a format usable by Julia. Instead, we want to allow
+  for roots to be discoverable at the beginning of a garbage collection.
+* *Conservative scanning of stack frames and objects.* Currently,
+  scanning does have to be precise. If we desire to use the GC for
+  foreign code that requires conservative scanning (especially for
+  foreign stack frames), then it is necessary to have functionality
+  that determines whether a machine word is a pointer to an object,
+  including to its interior.
```
yuyichao, line 40: It'll be better to have the three functions separated.

Custom layout and additional roots are useful features in general.
Conservative scanning is also fine, as long as it's clear it's up to the user to make sure their compiler doesn't mess things up and that it's up to the user to implement all the layouts supported by the julia GC.

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+```
+typedef void (*jl_gc_cb_pre_gc_t)(int full);
+typedef void (*jl_gc_cb_post_gc_t)(int full);
+typedef void (*jl_gc_cb_root_scanner_t)(int full);
+typedef void (*jl_gc_cb_task_scanner_t)(jl_task_t *task, int full);
+typedef void (*jl_gc_cb_external_alloc_t)(void *addr, size_t size);
+typedef void (*jl_gc_cb_external_free_t)(void *addr);
+
+#define jl_gc_register_callback(cb, func)
+#define jl_gc_deregister_callback(cb, func)
```
yuyichao: Implementation notes. It seems that there's no reason these should be macros. What they do can be implemented with functions without any problems. The GC code used to be full of such macros saving negligible typing and making the code very hard to reason about/search. These should be replaced with normal functions.

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+(`pre_gc`), the beginning of the mark phase (`root_scanner`),
+and the end of the GC (`post_gc`).
+
+As these callbacks are tested and called only once per collection, overhead
+should be negligible.
+
+Additional roots can be marked from the root scanner callback by
+calling the `jl_gc_mark_queue_obj()` function, which takes a pointer to
+the current thread's thread-local storage a pointer to the object as its
+parameters.
```
yuyichao: So this will mean that the scan of foreign object has to be done with breath first order, which will take much more space. If that's OK for the use case, it's fine. Otherwise, a much more integrated (and necessarily much less stable) API will be required.

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+
+```
+int jl_gc_mark_queue_obj(jl_ptls_t ptls, jl_value_t *obj);
```
yuyichao: Another implementation note, `gc_current_sp` should be just a member of the thread local storage.
Also, the place where `sp` could be updated externally is fairly well defined so it's better to just explicitly synchronize them at those point instead of leaking `&sp` from the beginning of the function.

It'll also make reasoning about work stealing between threads easier. (The marking is already thread safe AFAICT).

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+```
+
+*Performance impact:* Custom mark functions need to be called during the
+performance-critical mark loop of the garbage collector. In order to
+avoid overhead for the other cases, the code is engineered to consider
+such objects as the last possible option in the existing if-else chains.
+To accomplish that, such foreign types use the existing
+`jl_datatype_layout_t` structure, with `fielddesc_type` set to `3`,
+which is looked at after the other data types and the other alternatives
+for `fielddesc_type`.
```
yuyichao, line 227: (This is the "well defined point" where `sp` could be updated externally).

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+has been called for an old object and the mark function returns a
+non-zero value (thus showing how many young objects have been marked
+from the old one one), the GC knows to update its internal data
+structures accordingly.
+
+For an example of this, see the `gcext` test in the Julia repository,
+which defines a couple of such custom mark functions.
+
+To enable C finalization for such an object, the function
+`jl_gc_set_needs_foreign_finalizer()` has to be called for an
```
yuyichao: Why is this not implied/automatic?

GcExtensions.md
```diff
@@ -0,0 +1,353 @@
...
+In order to allow foreign code to have access to necessary functionality
+in the garbage collector, we allow foreign code to register callbacks for
+certain GC events. We provide for six types of callbacks:
+
+* `pre_gc` (beginning of garbage collection).
+* `post_gc` (end of garbage collection).
+* `root_scanner` (before scanning GC roots).
+* `task_scanner` (when scanning Julia tasks).
+* `notify_external_alloc` (when an external object is allocated).
+* `notify_external_free` (when an external object is freed).
```
yuyichao: Everyone of these can be called from any thread. Everyone other than `pre_gc` and `post_gc` can be called concurrently.

As for the current implementation, `post_gc` may not happen before the next `pre_gc`, which may or may not be what you want.

GcExtensions.md
rbehrends: Unfortunately, this cannot be done with functions in a typesafe fashion, as the type of the second argument depends on the value of the first. This is an attempt to work around the lack of generics in C.

That said, I'd be very much open for alternative solutions.

GcExtensions.md
fingolfin: What do you mean with that? The system must somehow be able to tell if finalizers should be run for a given object -- or perhaps rather the converse: we want to avoid the overhead for objects that don't need a finalizer.

I am probably missing something, could you perhaps elaborate? Thanks

GcExtensions.md
rbehrends: One option I have considered for this situation is to expose an interface to object arrays, which would handle that, but only works for object layouts where pointers are stored in consecutive locations.

GcExtensions.md
rbehrends: That is something that I'd very much like to do. The reason it's done in this roundabout way is that I didn't want to change the layout of the thread local storage for now (for reasons of stability and to keep the patch small, as it has been rebased on the most recent top of master repeatedly).

GcExtensions.md
rbehrends: The reason is that one may not want to finalize all objects of a given foreign type.

GcExtensions.md
yuyichao: Specialized registration function for each or use `void*` as function pointer argument type.

GcExtensions.md
yuyichao: Adding member to the end of tls isn't breaking. (No one should ever allocate that on their own)

GcExtensions.md
rbehrends: I'm aware of that (well, not the part that `post_gc` may happen before the next `pre_gc`). But yes, general implementations would have to be thread-safe. I will amend the text to make a note of that.

GcExtensions.md
fingolfin: They are already mostly separate, I'd say. At least I am not aware of anything in the API that would preclude you from only using one, or two, of them. If you see issues, we'd appreciate if you could give us a hint as to where. Thanks!

GcExtensions.md
yuyichao: K, sure, if there's usecase for that. Otherwise I think it's less error prone to just define a special function to allocate foreign object with the finalizer automatically added.

GcExtensions.md
yuyichao: Also, the worst part of this is making thing much harder to search, i.e. with the symbol concatenation. A macro with explicit spelling is also OK (same goes for the few new macros in `gc.c`).

fingolfin: Using `void *` would loose the type safety -- but I'd be fine with that, too (it's just another very small convenience for detecting very basic errors, but hardly important). And of course you are right, we could replace them by 2*6 custom (de)register functions, one for each hook -- I think the main annoying bit about that is the code duplication -- but if that's what you guys prefer, we could do it. We'll wait for some more feedback (perhaps some other Julia devs also have comments) and then look into revising the proposal. Thanks for your comments so far!

GcExtensions.md
yuyichao: It's not about this julep. I just mean that adding each of these separately is better, both for the commit (history) and for PR.

GcExtensions.md
fingolfin: Good news then! As @rbehrends said, we mostly were concerned about making a change that would be rejected out of hand. We'd certainly prefer `sp` in the TLS, too.

GcExtensions.md
rbehrends: Specialized registration functions make the interface very cumbersome, but it could be done. `void *` is not an option, as function pointers cannot be safely cast to `void *` (they can be safely cast to an arbitrary function pointer type and back, but that makes client code very verbose). And in either case, we would lose type safety. The way to go in a typesafe fashion would be to have separate registration/deregistration functions for all callbacks.

GcExtensions.md
yuyichao: Finallizers can run arbitrary julia code, make arbitrary allocations and must be run with the GC enabled. That's why multiple GC can happen and finish before hitting the `post_gc` callback, **even with one thread**. If that's not what you want, you may need to use another list instead of the normal finalizer run list.

GcExtensions.md
fingolfin: Ahhh, OK. So you are saying that if/when we submit our code as a PR, we should have this in three separate commits. I guess we could do that, gotta think about it (but again, we'll wait a bit for the dust to settle ;-).

GcExtensions.md
fingolfin: That's actually totally fine by us. Our finalizers right now are plain C functions which take great care to not trigger another GC in any way.

yuyichao: Other finalizers could still trigger GC so the ordering will still not be guaranteed. If that's OK, just documening the behavior should be fine.

GcExtensions.md
fingolfin: Unfortunately, this can change over the lifetime of an object -- i.e. an object may be created w/o needing a finalizer, but then later on it turns out that it needs one.

Just like it can happen with other Julia objects, via `jl_gc_add_finalizer`... And in fact, note that this is currently the implementation of `jl_gc_set_needs_foreign_finalizer`:
```C
JL_DLLEXPORT void jl_gc_set_needs_foreign_finalizer(jl_value_t *obj)
 {
     jl_gc_add_finalizer((jl_value_t *)obj, jl_nothing);
 }
 ```

**Comment by Keno** [comment]
I haven't read this through in detail, but presumably you need a mechanism to record the current state of all big objects at the moment you turn on conservative stack scanning to make sure you don't miss any objects in your binary tree.

Another thing to watch out for is what happens when whole pages get dropped at a time. In particular you need to
1) Make sure not to consider objects in dropped pages
2) Perhaps re-zero a page that's put back into service after being dropped to make sure you don't have stale data that could get fed into the GC.

The compiler writer in me also continues to be concerned about the proposition that conservative stack scanning is sound. There's quite a bit of fire you're playing with there, especially as compilers continue to improve.

**Comment by yuyichao** [comment]
> The compiler writer in me also continues to be concerned about the proposition that conservative stack scanning is sound. There's quite a bit of fire you're playing with there, especially as compilers continue to improve.

That's why I'm only fine with this since the conservative scan is all external.

> In particular you need to

There are already code in `gc-debug` that does this. I didn't check carefully if it's used/if it's used correctly since that's also completely a user problem.

**Comment by rbehrends** [comment]
We do indeed need to record bigval allocations; this is what the `notify_external_alloc` and `notify_external_free` hooks are for.

I originally believed that we could get away without zeroing pages by having sufficiently exhaustive validity checks for object data, but your comment made me revisit that assumption, and you are right. While pages that are currently dropped already aren't considered, the code did not properly account for pages being reused.

I have added the necessary code to fix that oversight. Because that introduced some minor, but non-zero overhead, I made zeroing contingent on a flag (enabled by `jl_gc_enable_conservative_scanning()`) in order to avoid that overhead if conservative scanning features are not needed.

I am aware of how optimizing compilers can create issues for conservative garbage collectors. This is discussed in some detail in Boehm, H., and D. Chase, "A Proposal for Garbage-Collector-Safe C Compilation", Journal of C Language Translation 4, 2 (December 1992), pp. 126-141, for example. There are also other concerns, such as integers being falsely interpreted as spurious stack roots on 32-bit systems with large heaps.

The assumption that we make here is that conservative stack scanning is rarely a desired choice, but something that is typically forced upon us, often by the use of or need to interoperate with legacy or third-party code or by overriding engineering concerns. It's something that we have to live it, which is why the proposal only has it as an opt-in mechanism on an as-needed basis rather than a default assumption.  So, yes, this is not an ideal solution, but rather a (sometimes) necessary one.

For example, one use case that we specifically require it for is interoperability with GAP, which assumes conservative stack scanning throughout, including many packages authored by external developers.

We do assume that compilers will not introduce changes (such as pointer masking) that would break the assumptions underlying conservative stack scanning. The reason is that too much existing code relies on conservative garbage collection (including, but not limited to any software that uses the Boehm GC) and such changes would break a lot of existing software. Therefore there exists a very strong disincentive for the major compiler vendors to introduce such changes (at least not without offering workarounds).

**Comment by Keno** [comment]
> We do indeed need to record bigval allocations; this is what the notify_external_alloc and notify_external_free hooks are for.

That's not quite what I meant. I mean that you need some mechanism of retrieving all such allocations that were created prior to installation of these hooks.

**Comment by Keno** [comment]
Forgot to mention. You'll also need a mechanism to scan all threads' registers (including x/y/zmm).

**Comment by Keno** [comment]
> We do assume that compilers will not introduce changes (such as pointer masking) that would break the assumptions underlying conservative stack scanning. The reason is that too much existing code relies on conservative garbage collection (including, but not limited to any software that uses the Boehm GC) and such changes would break a lot of existing software. Therefore there exists a very strong disincentive for the major compiler vendors to introduce such changes (at least not without offering workarounds).

I'm gonna stop talking about this, since I doubt I'll be able to change your mind, but I think you're underestimating what compiler writers are willing to do here. In any case, the only point I'm making here is that I continue to encourage you think think about ways to regain precision with your gap integration, even once you get this working.

**Comment by rbehrends** [comment]
> That's not quite what I meant. I mean that you need some mechanism of retrieving all such allocations that were created prior to installation of these hooks.

Oh, I see. Keep in mind that for conservative scanning we only need to scan foreign stack frames and we do it because we do not know their layout. But we can control what types of objects go into them. So, for the common use case that a foreign module needs to be able to find the objects that it creates within its stack frames, you can simply start tracking those after module initialization. Objects passed in from a Julia call should still be handled via `JL_GC_PUSH1()` and friends (or stored permanently inside other objects or roots).

If, for some reason, you want or need to track *all* Julia objects, then you do indeed have to set the callbacks before initialization. The alternative would be to be able to traverse all the existing bigval lists (one per thread), which would be somewhat invasive if it's to be done in a thread-safe fashion, and I'm not sure if it's worth it.

That said, if people think that there's a good use case or a need for it, I'd definitely be happy to add such a feature.

> Forgot to mention. You'll also need a mechanism to scan all threads' registers (including x/y/zmm).

Threads are tricky, but that's more because finding the end of the stack efficiently is hard, as Julia uses `sigaltstack()` and signals for its GC safepoint mechanism. I don't have the time tonight to do a full writeup for that (will do that later), but it basically comes down to writing some low-level code (some of it possibly architecture specific). I do not think at this point that much of this can really be alleviated much from the Julia side of things.

**Comment by fingolfin** [comment]
@Keno there really is no way (nor any need) to "convince" us regarding conservative stack scanning -- not because we prefer it (we are agnostic about this), or we don't want to try to be more precise (again: completely neutral, willing to try whatever works) -- but simply because it's a factual reality for us: the GAP code base is there, and in active use, we can't just drop it. It is also about as old as Python, and GAP 4 (the current major version) was started in 1996. Its C kernel consists of >100k SLOC of C code (plus a bunch of kernel extensions written by 3rd parties that need to stay compatible) -- perhaps not that big in some people's eyes, but certainly big enough so that "let's just rewrite it to use precise GC" is not something we can pull of easily. Esp. as we have to do this from inside academia, with little to no funding for thi.

As such, conservative scanning is not something we "want", or are particular in love with, it's just something that is there, in a decades old code base, deployed on dozens of different architectures and compilers in the past 30 years, with great success, and so far no issues with our GC anywhere.  And while of course tomorrow a compiler might be released which breaks it -- and we'll have to deal with that when/if it happens (but since older compiler versions will still be around, we'd still have a couple years for it, luckily). But still, that's easier said than done.

So we are in no rush to rewrite our battle tested conservative GC, or rewrite our whole code base from scratch (which we can't afford money wise, nor in terms of the countless bugs a full rewrite always introduces)... But we *do* think about this, and have a rough long-term plan to switch to a new GC... Like, e.g., the Julia GC... *hint* *hint*.

So, IMHO the better view point is this: Integrating GAP with the Julia GC actually could open up a way to slowly and incrementally migrate the GAP code base to not rely on a conservative scanner. So, you don't need to convince us about the pros of precise scanning vs. conservative, or about the dangers of crazy compiler writers, or any of that; just trust that we know what we are doing, just like we trust that you do, and consider it as you guys helping us to have a chance to one day bask in the glorious light provided by your wondrous precise garbage collector ;-).

In a sense, compilers breaking our conservative scanning would even help us (the people working on integrating the Julia GC with GAP), by making it way easier to market our effort to more GAP devs, many of whom remain sceptical about it (like: "will this upstart Julia project be still around in a few years? Should we really tie ourselves to it?" You gotta understand the graybeards, they saw so many other project come and go... Also: we take a 20% performance hit with Julia GC compared to our existing GC, though if/when we ultimately manage to combine multi threaded Julia with multi threaded GAP properly, this likely will change in favor for Julia's GC -- and we also have some ideas for further optimizations... Doing all that is for a later stage of this project, though... baby steps first ;-)

Anyway, all that said: we'd really love it if this proposal and the corresponding pull request were ultimately accepted (of course only after taking all your helpful feedback into account), so that we can eventually steal^H^H^H^H^H benefit from your great work in multiple ways: a shiny precise GC, a shiny fast and powerful language we can use to *incrementally* migrate parts of our kernel away from C (thus making it automatically "precise"), etc.

**Comment by Keno** [comment]
> Oh, I see. Keep in mind that for conservative scanning we only need to scan foreign stack frames and we do it because we do not know their layout. But we can control what types of objects go into them. So, for the common use case that a foreign module needs to be able to find the objects that it creates within its stack frames, you can simply start tracking those after module initialization. Objects passed in from a Julia call should still be handled via JL_GC_PUSH1() and friends (or stored permanently inside other objects or roots).

Ok, if you're willing to guarantee that no non-GAP julia bigval object will ever end up needing to be conservatively scanned, then tracking those objects is not necessary. That restriction would be a good thing to document in the interface though.

**Comment by Keno** [comment]
> Threads are tricky, but that's more because finding the end of the stack efficiently is hard, as Julia uses sigaltstack() and signals for its GC safepoint mechanism. I don't have the time tonight to do a full writeup for that (will do that later), but it basically comes down to writing some low-level code (some of it possibly architecture specific). I do not think at this point that much of this can really be alleviated much from the Julia side of things.

Well, it at the very least needs to be integrated with the julia safepoint mechansim such that you get the register contents.

**Comment by Keno** [comment]
Oh, @vtjnash and I were both concerned about write barriers, but I see you have them in the code. Would be good to document that requirement also.

**Comment by Keno** [comment]
@vtjnash points out that by keeping the appropriate entry in our finalizer list, it is eligible for early finalization (using the `finalize` function), which may not be what you want.

**Comment by vtjnash** [comment]
I think this is sounding good from our end.

One alternative approach could be to avoid finalizers and the jl_gc_cb_external_free_t callback altogether, and instead provide a hook that runs after marking, before sweeping. For example, I am using this type of hook to provide management of an external pool of stacks in one of my PRs: Doing the sweep at https://github.com/JuliaLang/julia/pull/13099/files#diff-6c07ad09c7451fb0b61b494536691bfaR148, and called from https://github.com/JuliaLang/julia/pull/13099/files#diff-31b93f4272eaeede3738531e90317e0cR2555. I don't think I actually have a preference either direction.

**Comment by rbehrends** [comment]
I have revised the proposal and its implementation in line with the suggestions made in this thread.

Major changes:

Commits have been refactored to separate changes according to their purpose.

The callback functions now avoid using macros. In order to not pollute the namespace unnecessarily, both adding and removing a callback is done using the same function, controlled by a flag.

C finalizers were reimplemented using the different technique suggested above by @vtjnash, avoiding some of the problems with the original version. We now call them sweep functions to make it clearer that they are more limited in nature than Julia finalizers and are meant for C-level resource and memory cleanup (fully featured finalizers can, after all, still be implemented in Julia and call out to C as needed).

Thread-local data was added to the end of Julia thread-local storage rather than in global arrays indexed by the thread id.

One can now also mark an object array in order to avoid mark stack explosion for custom mark functions of large objects.

Various functions were renamed to document their purpose more clearly.

Notes about thread-safety were added.

The text now stresses more clearly that the conservative scanning facilities are meant as a feature of last resort and should not be used indiscriminately.

Bug fixes:

Enabling the conservative scanning facilities is done using a separate function, which can be called both before or after `jl_init()`, so that the features for identifying objects in dropped and reused pages function correctly. We document that this may incur minor overhead (depending on how much the size of the heap fluctuates, but typically no more than 2% in benchmarks for heaps with regular fluctuations). This overhead is not incurred unless the feature is enabled.

File: GcExtensions.md (hunk 1 above)
Comment (fingolfin) [reply GcExtensions.md:40]: Note that this split has now been done, I hope it's now the way you wanted it.

**Comment by StefanKarpinski** [comment]
Shall we merge this Julep then? If @rbehrends feels that it's ready and @yuyichao, @vtjnash and @Keno agree.

**Comment by rbehrends** [comment]
From my side, it looks ready. Obviously, I don't know if you folks are happy with the interface, but I think I've at least addressed all concerns mentioned so far.

On the implementation side, everything looks fine to me, too, though I plan on tidying up the code a bit more and to rebase it onto a more recent version over the next few days.

**Comment by fingolfin** [comment]
Assuming this Julep gets accepted, what are the next steps? We open a PR on https://github.com/JuliaLang/julia/ with our implementation? Which branch should that target, master? (We are kinda hoping that this might get into Julia 1.1, but of course that depends on many factors, so I am not asking for anything here, just expressing a vague hope ;-)

File: GcExtensions.md [hunk 8]
```diff
@@ -0,0 +1,429 @@
+# Garbage Collector Extensions
+
+- **Title:** Garbage collector extensions for better foreign language support
+- **Author:** Reimer Behrends <behrends@gmail.com>
+- **Created:** May 2018
+- **Status:** work in progress
```
vtjnash, line 6: Can change this to `Accepted`, or we can wait for the corresponding PR (in case there are any changes)

GcExtensions.md
```diff
@@ -0,0 +1,429 @@
...
+typedef void (*jl_gc_cb_notify_external_free_t)(void *addr);
+```
+
+The allocation callback is invoked with the address and size of the
+new object, the deallocation callback is invoked with the address of
+the object about to be freed. Allocation and deallocation is still
+managed by Julia. The intent here is that foreign code can track
+allocations and deallocations in a data structure of its own if
+needed. An example of this can be seen in the `gcext` test, where
+we use a balanced tree to track allocations.
```
vtjnash, line 339: It might be nice just to move this logic into Julia and drop these hooks. Most of this tree already exists anyways (since we need it for pool allocations), and I think it might turn out to be useful to have. Anyways, that change can be done at any time.

vtjnash:
> We open a PR on https://github.com/JuliaLang/julia/ with our implementation? Which branch should that target, master? We are kinda hoping that this might get into Julia 1.1.

Yes, correct, and yes. The Julep sounds very straightforward – I don't foresee any problems with this getting merged fairly quickly.

GcExtensions.md
fingolfin: To clarify my understanding of what you wrote here: these notifications are for tracking malloc'ed memory ranges; in particular, to allow to take an arbitrary address and find out if it is inside a memory range we malloc'ed (and to find the start address of that range).  Are you saying this is something you might want to do inside Julia itself? If course we wouldn't mind that, as it'd simplify our code on the other side. We just didn't expect you to like the idea, for performance reasons (the bookkeeping is cheap but not free)

vtjnash: Yes, it seems like it might be useful – and could be enabled/disabled with a runtime function call (just as the callbacks already would be) to avoid introducing overhead when it isn't being used

rbehrends: It would not per se be difficult to reintroduce it; my original code in fact did that directly by using the `prev` and `next` references in `bigval_t` as left and right references for trees. However, this becomes less trivial once you introduce threads (the `gcext` test is single-threaded and therefore can avoid dealing with that issue). You either introduce a potential serialization bottleneck or you have to store them in thread-local storage and then deal with that split during garbage collection. By keeping that on the client side, we can (if necessary) adjust the implementation to our needs.

Also, currently we use treaps, because they're very simple to implement and verify (unlike red-black trees) and support efficient merge operations exist (for threaded use), but they have the downside that they are probabilistic data structures. This may require a reimplementation using a different form of balanced trees in the future.

In short, I'm not adverse of having that functionality available in Julia proper, but I think that should not happen without resolving these concerns first.

